PR #41, opened 2024-01-23
Add generation evaluation decorator

### LINKED ISSUE #39: add generation evaluation

vkehfdl1:
close #39

You can use this as decorator to any function that returns List string as generated answer, 
or string list, tokens (2-d list of int list), log_probs (2-d list of float list), as mentioned in gunn structure.


Eastsidegunn (approved):
LGTM
Later, I think we need to provide yeild return for token, log-probs scale evaluation which are used in high-advanced skill for example FLARE etc.
Even if it's not necessarily 'this' module.

PR #45, opened 2024-01-23
add global logger at __init__.py with RichHandler

### LINKED ISSUE #42: implement logging
make standard logging method for AutoRAG project

vkehfdl1:
close #42

I add global logger at autorag/__init__.py. 
But, when you use it, do not import directly from autorag/__init__.py because it will potentially occur circular import error.
Instead, you can get logger like below code. Remember logger instance have same name is singleton.
```python
import logging
logger = logging.getLogger("AutoRAG")
```

I use RichHandler instead of StreamHandler for beautiful formatting. 
Plus, I use rich exception for beautiful inspection while debugging.
You can see rich logging by deleting last two lines of your pytest.ini
Change pytest.ini looks like this.

```
[pytest]
env =
    OPENAI_API_KEY=sk-xxxxxxxxxxxxxxxx

log_cli=true
log_cli_level=INFO
```

### DIFF
--- a/autorag/__init__.py
+++ b/autorag/__init__.py
@@ -1,5 +1,11 @@
 __version__ = '0.0.1'
 
+import logging
+import logging.config
+import sys
+
+from rich.logging import RichHandler
+
 from .evaluator import Evaluator
 
 from llama_index import OpenAIEmbedding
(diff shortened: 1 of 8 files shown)


bwook00 (approved):
LGTM

Eastsidegunn (approved):
LGTM.
[rich handler docs](https://rich.readthedocs.io/en/stable/reference/logging.html#rich.logging.RichHandler)

PR #47, opened 2024-01-24
Change summary csv file to parquet file

### LINKED ISSUE #46: Refactor method that save best module and best param
Now, we save which module and param is best to the best filename directly.
But, it will occur some unintended errors when param or module name has underscore (_) or hyphen (-).
So, I have to find a way how to notify which module and module params are selected in node run function.

vkehfdl1:
close #46

- add metadatas at strategy.py `select_best_average` and `filter_by_threshold`. Now, that functions can retrieve filtered or selected metadata, which is corresponding with result inputs.
This feature is effective to preserve module's name and module's params when selecting best module & params with strategy.
- change all summary.csv to parquet files. csv file can't store dict params as dict type.
- add module_name, module_params at summary.parquet files. 
It was hard to classify underscore that divide param key and value, or underscore in param's key itself.
Now, we store dict itself to summary.parquet, so it will be no error or malfunctions.
- add 'is_best' column to node summary.parquet, so you can find best module name and module params easily with only look up node summary.parquet file.

### DIFF
--- a/autorag/node_line.py
+++ b/autorag/node_line.py
@@ -43,7 +43,14 @@ def run_node_line(nodes: List[Node],
     summary_lst = []
     for node in nodes:
         previous_result = node.run(previous_result, node_line_dir)
-        best_module_filename = os.path.basename(find_best_result_path(os.path.join(node_line_dir, node.node_type)))
-        summary_lst.append({'node_type': node.node_type, 'best_module_filename': best_module_filename})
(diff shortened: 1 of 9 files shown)


autorag/node_line.py
Eastsidegunn, line 1: Is best_execution_time most low value?

Eastsidegunn (approved):
LGTM.

autorag/node_line.py
vkehfdl1: Make this new issue thx

bwook00 (approved):
LGTM

PR #74, opened 2024-01-29
Add evaluation retrieval contents metric and its evaluation functions

### LINKED ISSUE #73: Add Passage contents - Retrieved_contents relevancy metrics
Add some metrics for evaluating context relevancy.
It will calculate how retrieved passage contents are similar with retrieval gt contents. 

It will have crucial part for evaluating passage compressor node.

vkehfdl1:
close #73

I need to calculate retrieval contents token f1 score for evaluating passage compressor.

### DIFF
--- a/autorag/evaluate/metric/retrieval_contents.py
+++ b/autorag/evaluate/metric/retrieval_contents.py
@@ -43,3 +43,19 @@ def retrieval_token_f1(gt: List[str], pred: List[str]):
     _, _, result = zip(*calculated_results)
     result_np = np.array(list(result)).reshape(len(pred), -1)
     return result_np.max(axis=1).mean()
+
+
+@retrieval_contents_metric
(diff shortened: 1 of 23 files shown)


autorag/evaluate/metric/retrieval_contents.py
Eastsidegunn, line 1: I think for the passage compressor, 
compress ratio can be an option for your task.
semantic metric needs to be considered, but there is no clear approach.
n-gram based metric can only present original content conservation. in my think... LGTM cheer up!

bwook00 (approved):
LGTM

autorag/evaluate/metric/retrieval_contents.py
vkehfdl1: I agree. There can be another great strategy to evaluate it. Thx:)

PR #96, opened 2024-02-01
change module result filename to numbers

### LINKED ISSUE #86: Don't put module params in make module file name

vkehfdl1:
close #86

- delete `make_module_file_name` function
- convert module result filenames to number parquet file, like 0.parquet.

This feature will prevent malfunction due to long module param name, or special letter like slash (/) or underscore(_) and so on.
We can still keep our feature because we store module_params as dict at summary.csv


bwook00 (approved):
LGTM

Eastsidegunn (approved):
LGTE

PR #98, opened 2024-02-01
Make Hybrid Retrieval RRF module and edit run.py for it

### LINKED ISSUE #82: Add Hybrid Retrieval Module
There will be possible many things to edit at retrieval run.py

vkehfdl1:
close #82

<Things that I developed>
1. Add Hybrid RRF module.
Reciprocal Rank fusion the results of retrievals.
It uses ids and scores from `already-done` results.

To use Hybrid module, you need to do 2 things at your yaml file.
First, you need to set more than 2 retrieval modules. 
Second, set names of your module as `target_modules` at yaml file as tuple.

2. Add tuple function at yaml loader.
Initially, tuple does not support at yaml file.
But, there was no way to pass multiple value parameters at module functions in AutoRAG, due to `make_combinations` function. It is function that make all combinations inside module params list. 
So, I make tuple at config yaml file, and turn it to *real* tuple using `convert_string_to_tuple_in_dict` function.
Now, you can load tuple as single module parameter.

3. Execution time of hybrid retrieval.
Hybrid Retrieval module in AutoRAG did not `acutally` run each modules, so it can't get *real* execution time. 
So, I extract best module from each target_modules, and get their exectuion time and sum it.
This way, we can properly filter by speed (execution_time) at hybrid retrieval.

4. Fix error at `make_combinations` at delete duplication.
As you see at number 2, tuple occured error at `list(set())` for deleting duplication at module params.
So, from now, we do not perform duplication delete at unhashable objects like tuple, list, dict..
Most of time, the duplication deletion feature will perform.


bwook00 (approved):
LGTM

autorag/nodes/retrieval/hybrid_rrf.py
Eastsidegunn, line 1: in my idea, a hybrid mechanism like this is different from other retrieval modules.
so the difference need to be described in the annotation.
this is special case.

vkehfdl1: I will explain more about hybrid function at documentation later.

Eastsidegunn (approved):
LGTM :>

Eastsidegunn:
have a nice night

bwook00 (approved):
LGTM

PR #103, opened 2024-02-03
Fix errors at Runner for deploy feature

### LINKED ISSUE #83: potentially Runner.run function will not work.
It override previous result, so delete all columns except result column.
So, it will can't preserve older results, it can be cause malfunction at certain RAG workflows.

vkehfdl1:
close #83

- add full trial folder contains result for testing full deploy function from real result. It added many files, so just ignore numerous 'file changed'
- add project_dir at `Runner` initialize and `from_trial_folder` for user experience. When you run `Runner.from_trial_folder`, you don't have to run that code from the project directory.
- Sort node line order by config.yaml file. It has error that extracted yaml file from trial folder can be mixed when there are multiple node lines.
- Concat previous result and newest result, and passed it. In `run.py` this feature supported, but not supported at Runner. So I made it.


Eastsidegunn (approved):
LGTE

bwook00 (approved):
LGTM

Eastsidegunn (approved):

bwook00 (approved):
LGTM

PR #105, opened 2024-02-03
add embedding_models and generator_models

### LINKED ISSUE #60: Add Embedding models

### LINKED ISSUE #104: Add LLM Models

bwook00:
close #60 
close #104


Eastsidegunn (approved):

vkehfdl1 (approved):
LGTM
As you know, since we not initialize llm instance, there can never be problem about import.
Thus, changing embedding_models value to class itself, not instance, might be better than this way in the future.

PR #107, opened 2024-02-03
Enable using hybrid retrieval at deploy.

### LINKED ISSUE #91: Add Hybrid Retrieval "Deploy" Function.
It is so confusing and difficult to make deploy function at retrieval_node for hybrid retrieval.
So, I have to make deploy function for hybrid retrieval.
Plus, I have to make hybrid retrieval yaml file for deployment.

vkehfdl1:
The original `hybrid_rrf` and `hybrid_cc` functions takes `ids` and `scores` as their input. So, it was hard to use these hybrid modules at deploy.
To understand this difference, you must understand the way optimization and deploy `run` is working.
As you know, the module functions (including its decorator of course) is treat as independent functions both on optimization and deployment process. 
In optimization process, we rely `run.py` functions for executing each module functions. Most of the case, `run.py` do not run something special for **running modules**, but hybrid was special. To running hybrid functions, retrieval `run.py` must select best modules among user's `target_modules` input, and extracting its ids and scores. Finally, we could use hybrid modules properly.
However, in deployment `Runner.run`, do not use `run.py` functions. Because `run.py` functions merely contains optimization process, and it is super inefficient to use `run.py` at deployment feature.
Instead using `run.py`, at `Runner` extracts the best module name and module parameters from `summary.csv`, which made at `run.py`, and construct new config dictionary. With that dictionary, `Runner` can run whole modules one by one with selected parameters.

And here was the problem. As you know, hybrid modules must pass `ids` and `scores`, and that parameters made at `run.py`. But, at summary.csv, there was no `ids` and `scores` parameters, but `target_modules`, because summary.csv save user's input parameters as default. Hence, it was impossible to run hybrid modules at deployment `Runner` class.

So, what was the solution? 
I swapped module parameters of hybrid modules at `summary.csv` in retrieval `run.py`. It means, I delete `ids`and `scores` at module params, and add `target_modules` and `target_module_params` as new hybrid module params.
And at the retrieval decorator, if there are no `ids` and `scores` parameters, I run other retrieval module with input `target_modules` and `target_module_params`. 
In this way, you can run another retrieval module at retrieval node decorator, and obtain `ids` and `scores`, which is input of hybrid modules.
Since summary.csv module params saved with `target_modules` and `target_module_params`, we now can use hybrid modules at deployment!!

close #91

p.s. I thought it was great challenge for me, but it resolved pretty simple. I think the isolation of three parts (optimization, deployment runner, and modules) is really great and flexible. Maybe we can find a way to resolve some other weird methods, thanks to this isolation structure.


Eastsidegunn (approved):
LGTM

bwook00 (approved):
LGTM

PR #111, opened 2024-02-03
add sample data eli5, msmarco, triviaqa

### LINKED ISSUE #108: upload sample qa, corpus data

bwook00:
close #108


vkehfdl1 (changes requested):
Why don't you change all filenames to `qa.parquet` or `corpus.parquet`
We use strict filename rule for this dataset, and it will be easy that users just copy&paste that datasets, without rename it.
Plus, users already recognize which dataset is asqa or eli5 easily, because you separated to different folder.

vkehfdl1:
Plus, edit dataset as the real structure. This data must be the perfect example for evaluation dataset. It must be followed our rules for dataset rules.

sample_dataset/eli5/load_eli5_dataset.py
```diff
@@ -0,0 +1,29 @@
...
+    # set file path
+    file_path = "MarkrAI/eli5_sample_autorag"
+
+    # load dataset
+    corpus_dataset = load_dataset(file_path, "corpus")['train'].to_pandas()
+    qa_dataset = load_dataset(file_path, "qa")['train'].to_pandas()
+
+    # path setting
+    sample_dataset_dir = pathlib.PurePath(__file__).parent.parent
+    project_dir = os.path.join(sample_dataset_dir, "eli5")
```
vkehfdl1, line 17: You can just save this to `pathlib.Purepath(__file__)` directory.
Because `__file__` is load_eli5_dataset.py file path, so you can save qa.parquet and corpus.parquet at that directory,

sample_dataset/README.md
```diff
@@ -0,0 +1,5 @@
+# sample_dataset handling
+
+The sample_dataset folder does not includes a `qa.parquet`, `corpus.parquet` file that is significantly large and cannot be uploaded directly to Git due to size limitations. 
+
+To manage this, we have developed Python script named `load_dataset.py` that can efficiently load this dataset file and put it in each project folder.
```
vkehfdl1, line 5: Add simple sample code snippet for downloading data.

vkehfdl1 (changes requested):

sample_dataset/README.md
```diff
@@ -0,0 +1,30 @@
...
+If you run the script without the `--save_path` argument, the dataset will be saved to a default location, which is the directory containing the `load_triviaqa_dataset.py` file, essentially `./sample_dataset/triviaqa/`:
+```bash
+python ./sample_dataset/triviaqa/load_triviaqa_dataset.py
+```
+This behavior allows for a straightforward execution without needing to specify a path, making it convenient for quick tests or when working directly within the target directory.
+
+### 3. Utilizing the if __name__ == "__main__" block
+If the Python file contains an `if __name__ == "__main__":` block, executing the file directly will trigger any code within this block. 
+
+This block can include function calls for loading or processing the dataset, which will execute when the file is run directly.
```
vkehfdl1, line 30: useless annotation
delete it.

vkehfdl1 (changes requested):

vkehfdl1 (approved):

Eastsidegunn (approved):
LGTM

PR #115, opened 2024-02-04
Fix VERSION file and change logging level to INFO

### LINKED ISSUE #112: [HotFix] Can't run AutoRAG when installed.
At __init__.py, we cannot find `VERSION` file because it is not loaded to pypi.

```python
root_path = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
version_path = os.path.join(root_path, 'VERSION')

with open(version_path, 'r') as f:
    __version__ = f.read().strip()
```

This is the problem of code. We must change the path of VERSION, or find another solution.

vkehfdl1:
close #112


Eastsidegunn (approved):
LGTM

bwook00 (approved):
LGTM

PR #118, opened 2024-02-04
Enable async operation at hyde and query decompose

### LINKED ISSUE #117: Use acomplete at hyde and query_decompose
Now, hyde and query decompose do not using `acomplete` for async operation.
So it is using asyncio, the async operation we want, the llm api call, do not run as async.
Fix this problem for faster processing.

vkehfdl1:
close #117

It saves a lot of time!


bwook00 (approved):
LGTM

Eastsidegunn (approved):
LGTM... That's all right?

PR #126, opened 2024-02-04
enable target_module_params at hybrid retrieval in run.py

vkehfdl1:
In this way, we now evaluate test data easily with extracted yaml file!


bwook00 (approved):
LGTM

Eastsidegunn (approved):

PR #129, opened 2024-02-05
[HotFix] do not join inner at hybrid_cc for consistent 'top_k' result. It can …

### LINKED ISSUE #128: [HotFix] Fix that hybrid_cc return exactly top_k.
If there is no duplication of retrieved_ids, hybrid_cc drop that row as default.
Because of that, we have empty retrieved_contents at hybrid_cc and it occurs error at rerank or others.

To solve this, we have to make methods to hedge this kind of error. Like fill the scores to 0 like hybrid_rrf.

vkehfdl1:
…prevent errors becuase of empty retrieved ids.

close #128


Eastsidegunn (approved):
LGTM for now

bwook00 (approved):
LGTM

PR #163, opened 2024-02-13
Find a better way to using swifter

### LINKED ISSUE #161: [HotFix] missing import swifter at metric/generation.py

vkehfdl1:
close #161

There was problem that we can't use Pycharm 'optimize import' feature because of swifter. 
It needs to be imported for using, while Pycharm recognize swifter is not using.

This occurs because, `import swifter` actually not using at that file. Here are the reasons.

### How swifter works?

At swifter __init__.py, there is a function for register swifter functions to pandas.

```python
if "modin.pandas" in sys.modules:
    register_modin()
```

So, when you import swifter, it runs `__init__.py` and run `register_modin()`. 
It register swifter package to pandas. Pandas has external api functions to register add-ons.

Because of that, without `import swifter` we couldn't use swifter. But once it is run at `AutoRAG` somewhere, we can use it.

### What I did

What I did is this.
First, import `swifter.set_defaults` for running swifter `__init__.py` (+fast processing for strings)
Second, remove `import swifter` at all files. 
Lastly, turn 'optimize import' feature again! 

It works because all AutoRAG functions must be run after highest `__init__.py` run. And it runs swifter `__init__.py`

So @Eastsidegunn @bwook00 
Don't forget to turn on 'optimize import' at commit option!!

FYI : I tested all test and it passed.

### DIFF
--- a/autorag/__init__.py
+++ b/autorag/__init__.py
@@ -1,19 +1,21 @@
-import os
 import logging
 import logging.config
+import os
 import sys
-
-from rich.logging import RichHandler
 
 from llama_index.embeddings import OpenAIEmbedding, HuggingFaceEmbedding
 from llama_index.embeddings.openai import OpenAIEmbeddingModelType
 from llama_index.llms import OpenAI, Anthropic, AzureOpenAI, HuggingFaceLLM, LangChainLLM, GradientBaseModelLLM, \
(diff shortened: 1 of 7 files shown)


bwook00 (approved):
LGTM

Eastsidegunn (approved):
Haha! I'm already turn off that option!

vkehfdl1:
> Haha! I already turn off that option!

@Eastsidegunn Now please turn on that option!
I said "Don't forget to turn on 'optimize import' at commit option!!"

PR #169, opened 2024-02-15
Add env variable at config YAML file

### LINKED ISSUE #131: Add env variable for easy-to-use
When user needs to write their api key or api base directly to yaml, it is much better to use env varaible.
But, yaml doesn't support env variable as default.
So, it will be great we support env variable at config yaml file.

Below is the phind answer for making env varaible in yaml file.

---

To use environment variables inside a YAML file when loading it with PyYAML's `safe_load()` function, you need to add custom logic to find and resolve these environment variables. This involves defining a pattern to identify environment variables, creating a custom constructor to replace placeholders with actual environment variable values, and registering this constructor with PyYAML.

Here's a step-by-step guide based on the provided sources:

1. **Define a Pattern to Identify Environment Variables**: Use a regular expression to match your environment variable placeholders. For example, if your placeholders look like `${ENV_VAR}`, you can use the following pattern:
   ```python
   import re
   env_pattern = re.compile(r".*?\${(.*?)}.*?")
   ```

2. **Create a Custom Constructor**: This function will be called by PyYAML when it encounters a node that matches the pattern defined in the previous step. It should replace the placeholder with the actual value of the environment variable:
   ```python
   import os
   def env_constructor(loader, node):
       value = loader.construct_scalar(node)
       matches = env_pattern.findall(value)
       for match in matches:
           value = value.replace(f"${{{match}}}", os.environ.get(match, ""))
       return value
   ```

3. **Register the Pattern and Constructor with PyYAML**: Before calling `safe_load()`, you need to tell PyYAML about your custom pattern and constructor. You also need to specify a custom tag (e.g., `!env`) to trigger your constructor:
   ```python
   import yaml
   yaml.add_implicit_resolver("!env", env_pattern)
   yaml.add_constructor("!env", env_constructor)
   ```

4. **Load Your YAML File**: Now, you can load your YAML file. Any occurrences of `${ENV_VAR}` will be replaced with the corresponding environment variable's value:
   ```python
   with open('your_file.yaml', 'r') as file:
       config = yaml.safe_load(file)
       print(config)
   ```

Here's an example of what your YAML file might look like:
```yaml
config:
  username: admin
  password: ${SERVICE_PASSWORD}
  service: https://${SERVICE_HOST}/service
```

And here's how you might set the environment variables in a Unix-like shell before running your Python script:
```shell
export SERVICE_PASSWORD=yourpassword
export SERVICE_HOST=yourhost.com
```

This approach allows you to dynamically configure your application based on the environment it's running in, making it easier to manage configurations across different deployment environments without hardcoding sensitive information into your codebase or YAML files.

vkehfdl1:
close #131

### DIFF
--- a/autorag/evaluator.py
+++ b/autorag/evaluator.py
@@ -19,7 +19,7 @@ from autorag.nodes.retrieval.vectordb import vectordb_ingest
 from autorag.schema import Node
 from autorag.schema.node import module_type_exists, extract_values_from_nodes
 from autorag.utils import cast_qa_dataset, cast_corpus_dataset
-from autorag.utils.util import load_summary_file, convert_string_to_tuple_in_dict
+from autorag.utils.util import load_summary_file, convert_string_to_tuple_in_dict, convert_env_in_dict
 
 logger = logging.getLogger("AutoRAG")
 
@@ -168,6 +168,7 @@ class Evaluator:
                 raise ValueError(f"YAML file {yaml_path} could not be loaded.") from exc
 
         yaml_dict = convert_string_to_tuple_in_dict(yaml_dict)
+        yaml_dict = convert_env_in_dict(yaml_dict)
         node_lines = yaml_dict['node_lines']
         node_line_dict = {}
         for node_line in node_lines:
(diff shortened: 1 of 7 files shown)


Eastsidegunn (approved):

bwook00 (approved):
LGTM

PR #194, opened 2024-02-22
Feature/#160

### LINKED ISSUE #160: Refactor all test codes for fast testing and auto testing
We disabled our test CI because of the slow speed plus storage limitation.
We have to optimize this.
For example, LlamaIndex is large and complex framework with hundreds of unit test, but it takes only 19 seconds to test all things.
However, ours was more than 5 minutes and it is now not working at default ubunutu runner in github.

The suggest workflow is like below.

1. Do not use real LLM. Just use `MockLLM` for mocking the functionality of LLM. 
2. Use MonkeyPatch to using LLM or local Model. Testing and donwloading whole model like TART or MonoT5 tooks long time and large download size.
3. Do not use real embedding model. Use Mock embedding model or monkey patch.

After making test faster and lighter, we will enable auto-testing with CI again.

vkehfdl1:
close #160

Change all openai llm call to MockLLM call in testing.


Eastsidegunn (approved):

bwook00 (approved):
LGTM

Issue #207, opened 2024-03-06
late init embedding model


vkehfdl1:
When embedding model initialization, it is likely to be downloaded all embedding models. 
It is super inefficient since someone don't want to use local embedding, or only one embedding.

[What to fix]
- All embedding model instances is initializing at autorag/__init__.py. Late initialization all embedding models.

[How to]
- Suggest to search about `python late init`

PR #208, opened 2024-03-06
Unallocate gpu-intensive models after use

### LINKED ISSUE #205: unallocate gpu CUDA memory after processing each reranker, metric model, embedding model
At intense optimization, lots of models allocate to gpu memory.
This happens in single process, so it is hard to orchestrate each models. 
So it must unallocate gpu cuda memory when each node(or module) is done.

vkehfdl1:
close #205

- I didn't implement embedding models yet. Will do at #207


bwook00 (approved):
LGTM

Eastsidegunn (approved):

PR #206, opened 2024-03-06
edit guidance requirements to github

vkehfdl1:
close #202

Issue #202, opened 2024-03-05
change guidance requriements to follow github main branch


vkehfdl1:
Due to the late deployment of guidance team + error about ‘tests’ folder that install guidance, it can be occur error when we use guidance pypi latest version.
The github main branch resolve this issue at their PR, so it will be great we follow guidance github version since they deploy current version to pypi.

PR #209, opened 2024-03-07
dump version v0.0.11

vkehfdl1:
- build docs
- Unfortunately, dependency from git cannot upload to pypi. Look at #202 and #206. We still have this issue.... But, there is no way to fix this until guidance team release their latest version to pypi or deleting guidance dependency from AutoRAG.


bwook00 (approved):
LGTM

Eastsidegunn (approved):

PR #212, opened 2024-03-07
Implement LazyInit embedding_model

### LINKED ISSUE #207: late init embedding model

bwook00:
close #207 

- Since `lateinit` is a concept used in kotlin, I used `LazyInit` (which I found to be the best way to do it)
-> By doing so, we can defer the allocation of resources until needed, reduce initial loading times, and optimize memory usage. 

- Additionally, I added `torch.cuda.empty_cache()` to optimize memory usage.


autorag/evaluator.py
```diff
@@ -128,6 +128,7 @@ def __embed(self, node_lines: Dict[str, List[Node]]):
                     raise KeyError(f"embedding_model_str {embedding_model_str} does not exist.")
                 vectordb_ingest(collection, self.corpus_data, embedding_model)
                 logger.info(f'VectorDB corpus embedding complete with {embedding_model_str}.')
+                del embedding_model
```
vkehfdl1, line 131: add cuda empty_cache at here??

vkehfdl1 (changes requested):
I want to empty gpu memory with empty cache. In case embedding uses cuda

autorag/evaluate/metric/generation.py
```diff
@@ -144,6 +145,10 @@ def sem_score(generation_gt: List[str], pred: str, embedding_model: Optional[Bas
     gt_embeddings = embedding_model.get_text_embedding_batch(generation_gt)
     pred_embedding = embedding_model.get_text_embedding(pred)
 
+    del embedding_model
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
```
vkehfdl1, line 150: Here is the issue.
As you see, this `sem_score` function will run so many times in `generation_metric` decorator. 
Then, the deletion of embedding_model at every `sem_score` function is super inefficient. Because of this, every row, the python must load all models from the start. It takes super long time for I/O.

[Suggest Solution]
Do not use generation_metric decorator at `sem_score`. Like meteor or rouge. 

Init embedding_model only one time. And delete it at the end, after all row evaluation is end.

vkehfdl1 (changes requested):

Eastsidegunn (approved):

Eastsidegunn (approved):

autorag/evaluate/metric/generation.py
```diff
@@ -124,29 +125,46 @@ async def compute(gt: List[str], pred: str) -> float:
     return result
 
 
-@generation_metric
-def sem_score(generation_gt: List[str], pred: str, embedding_model: Optional[BaseEmbedding] = None) -> float:
+def sem_score(generation_gt: List[List[str]], generations: List[str],
+              embedding_model: Optional[BaseEmbedding] = None,
+              batch: int = os.cpu_count()) -> List[float]:
```
vkehfdl1, line 130: I think default batch size is way more larger than cpu core count? 
Like 64 or 128? Because it uses CUDA, not cpu.

vkehfdl1 (changes requested):

autorag/evaluate/metric/generation.py
```diff
@@ -124,29 +125,43 @@ async def compute(gt: List[str], pred: str) -> float:
     return result
 
 
-@generation_metric
-def sem_score(generation_gt: List[str], pred: str, embedding_model: Optional[BaseEmbedding] = None) -> float:
+def sem_score(generation_gt: List[List[str]], generations: List[str],
+              embedding_model: Optional[BaseEmbedding] = None) -> List[float]:
```
vkehfdl1, line 129: ??? Where is the batch input...?
My review was "change default setup of batch", not delete as input parameter.

vkehfdl1 (changes requested):

autorag/evaluate/metric/generation.py
```diff
@@ -138,13 +139,15 @@ def sem_score(generation_gt: List[List[str]], generations: List[str],
     :param embedding_model: Embedding model to use for compute cosine similarity.
         Default is all-mpnet-base-v2 embedding model.
         The paper used this embedding model.
+    :param batch: The batch size for processing.
+        Default is your cpu count.
```
vkehfdl1, line 143: Default is 128

vkehfdl1 (changes requested):

vkehfdl1 (approved):
LGTM🤯

Issue #44, opened 2024-01-23
Add n-gram metric options


Eastsidegunn:
From PR #40 

I think these n-gram based metrics can open some options.
- N : int  minimum unit for n-gram
- tokenizer : select tokenizer. (according to language, method etc include in tokenizer, tokenizer can influence value of each metric)
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
- etc

---

- [x] #222
- [x] #223

vkehfdl1:
How can get and write option about metrics at yaml file? It can be tricky, but if it is necessary, we'll find a way.

vkehfdl1:
close this issue because all generation metric has options. 
DONE!

PR #213, opened 2024-03-07
Use native Rouge Scorer instead of huggingface evaluate + async operation + additional options

### LINKED ISSUE #177: batch processing at rouge_scorer
Rouge_scorer constantly pops up logging `Using default tokenizer.`
Plus it is quite slow.

vkehfdl1:
close #177

It partially include #44. 

- add n-gram and other options in rouge score
- async operation at rouge
- do not raise any warning(logging info) about 'using default tokenizer'

### DIFF
--- a/autorag/evaluate/metric/generation.py
+++ b/autorag/evaluate/metric/generation.py
@@ -1,3 +1,4 @@
+import asyncio
 import functools
 import os
 from typing import List, Optional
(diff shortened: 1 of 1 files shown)


Eastsidegunn (approved):

bwook00 (approved):
LGTM

PR #216, opened 2024-03-08
Add vllm Module at generator

### LINKED ISSUE #214: Add vllm 'llm module' for faster generation performance
LlamaIndex supports `vllm` model. However, it is too slow at batch processing. It literally do not support async inference.
Look at my issue at [here](https://github.com/run-llama/llama_index/issues/10496).
So, it does not support any `async` call and it is super slow. It is basically linear process.

I looked at both `llama_index` and `vllm` code, and I found this at vllm code.
```python
def generate(
        self,
        prompts: Optional[Union[str, List[str]]] = None,
        sampling_params: Optional[SamplingParams] = None,
        prompt_token_ids: Optional[List[List[int]]] = None,
        use_tqdm: bool = True,
        lora_request: Optional[LoRARequest] = None,
    ) -> List[RequestOutput]:
```
You can see that you can put multiple prompts at once. 
But look at `llama_index.complete()` method.
```python
    def complete(
        self, prompt: str, formatted: bool = False, **kwargs: Any
    ) -> CompletionResponse:
        kwargs = kwargs if kwargs else {}
        params = {**self._model_kwargs, **kwargs}

        from vllm import SamplingParams

        # build sampling parameters
        sampling_params = SamplingParams(**params)
        outputs = self._client.generate([prompt], sampling_params)
        return CompletionResponse(text=outputs[0].outputs[0].text)
```

Just using "only first" index of input list.

Think about any ML model. You can multiple inputs as batch, and it process async at "GPU" level, until it is not occur OOM error.
I will set proper batch and put to vllm instance directly, since there is **no way** to support real async operation of Vllm model. 
Because their abstraction methods only supports "single prompt" as input.

In my opinion, the optimization of processing speed itself is crucial to AutoRAG. Because it takes so....long time;; I had to optimize it.

vkehfdl1:
- add `vllm` module in the generator node
- input all prompts at once inside the vllm instance. It looks like it control prompts parallel processing itself. 
- Make vllm test file, simplify both vllm test code and llama_index llm test code.
- I tested in CUDA environment and it works great.

close #214

### DIFF
--- a/autorag/nodes/generator/__init__.py
+++ b/autorag/nodes/generator/__init__.py
@@ -1,1 +1,2 @@
 from .llama_index_llm import llama_index_llm
+from .vllm import vllm
--- a/autorag/nodes/generator/base.py
+++ b/autorag/nodes/generator/base.py
@@ -39,4 +39,10 @@ def generator_node(func):
             result = func(prompts=prompts, llm=llm_instance, batch=batch)
             del llm_instance
             return result
+        elif func.__name__ == 'vllm':
+            return func(prompts=prompts, llm=llm, **kwargs)
+        else:
+            raise ValueError(f"{func.__name__} is not a valid generator node name. "
+                             "Please check the generator node name.")
+
     return wrapper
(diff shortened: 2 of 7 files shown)


Eastsidegunn (approved):

bwook00 (approved):
LGTM

PR #220, opened 2024-03-12
limit max length of all-mpnet-base-v2 model to 512 for preventing err…

### LINKED ISSUE #215: Surpass context length when calculating sem_score.
[What is the problem?]
While using sem_score with default mp_net or other local model, you can face ‘AssertionError’ when running in CUDA.
It is because input sentence surpass its context length limitation. 
We have to prevent this kind of error.

[How to?]
I think there is a few method to prevent this. 
1. Max length option?
2. Auto-retry
3. 1+2
Or any other solution will be welcome. Choose wisely.

vkehfdl1:
…or at CUDA processing.

close #215

Please look at the description at the issue tab.
Actually LlamaIndex handles itself about token limit when embedding.
But all-mpnet-v2 models `config.json` was wrong. It does not working more than 512 token dimension.


bwook00 (approved):
LGTM

Eastsidegunn (approved):
Good

PR #236, opened 2024-03-22
Add duplicate id check in BM25 ingest.

bwook00:
Feature/#93 While solving restart point issue, I realized that BM25 ingest should also have duplicate id check function like vectorDB(Chroma).

- The existing `pickle.dump` doesn't update, it just pushes everything and rewrites it. So I changed the duplicate_id_check function to find new_ids and proceed with ingest only when new_ids are found.

+ In the duplicate_id_check, I had written code to just get all the ids in the pickle file and then check if any of them overlapped with the ids I wanted to put in. I realized that this was inefficient, so I changed it to use `.isin()` in the pandas dataframe to get only duplicate ids instead of fetching all the ids in pickle.


autorag/nodes/retrieval/bm25.py
```diff
@@ -94,18 +95,48 @@ def bm25_ingest(corpus_path: str, corpus_data: pd.DataFrame):
...
+    # Load the BM25 corpus if it exists and get the passage ids
+    if os.path.exists(corpus_path) and os.path.getsize(corpus_path) > 0:
+        with open(corpus_path, 'rb') as r:
+            corpus = pickle.load(r)
+            bm25_corpus = pd.DataFrame.from_dict(corpus)
+        stored_passage_rows = bm25_corpus[bm25_corpus['passage_id'].isin(ids)]
+        stored_passage_ids = stored_passage_rows['passage_id'].tolist()
+        # Filter ids and contents for those not existing in the BM25 corpus
+        new_ids = [passage_id for passage_id in ids if passage_id not in stored_passage_ids]
+        new_contents = [contents[i] for i, passage_id in enumerate(ids) if passage_id in new_ids]
```
vkehfdl1, line 114: Why don't you use filter method like this?
```python
new_ids = list(filter(lambda passage_id: passage_id not in stored_passage_ids, ids))
```

vkehfdl1: Or as alternative, you can just use `stored_passage_rows` directly. 
I think it is pd.DataFrame, right?
https://jimmy-ai.tistory.com/346

```python
new_passages = bm25_corpus[~bm25_corpus['passage_id'].isin(ids)]
```

This can be really simple. Suggested.

autorag/nodes/retrieval/bm25.py
```diff
@@ -94,18 +95,48 @@ def bm25_ingest(corpus_path: str, corpus_data: pd.DataFrame):
...
+
+    if new_ids:
+        tokenizer = AutoTokenizer.from_pretrained('gpt2', use_fast=False)
+        tasks = list(
+            map(lambda x: bm25_tokenize(x[0], x[1], tokenizer), zip(new_contents, new_ids)))
+        loop = asyncio.get_event_loop()
+        results = loop.run_until_complete(asyncio.gather(*tasks))
+        tokenized_corpus, passage_ids = zip(*results)
+
+        new_bm25_corpus = pd.DataFrame({
```
vkehfdl1, line 127: If you re-make `new_bm25_corpus` again, why did you use `new_ids` and `new_contents` separately?

First, make bm25.pkl dict to pd.DataFrame.
Second, do tokenize new ids. 
You can apply with async using `swifter`.
Third, concat it!

vkehfdl1 (changes requested):

vkehfdl1 (approved):
LGTM

Eastsidegunn (approved):
LGTM

PR #240, opened 2024-03-25
Resolve long token embedding error at openai

### LINKED ISSUE #113: Parallel vectordb embedding
Use llamaindex `aget_text_embedding_batch` for faster ingestion.
The default implementation is simple loop, and it is inefficient.

### LINKED ISSUE #237: OpenAI embedding surpass token limit error
When you try to embed over 8192 tokens, there is an error at openai embedding. 
Huggingface embedding manage itself automatically its token size, but maybe openai embedding does not do that.

It can resolve in our codebase temporaly, but I think we can contribute to llama index.

vkehfdl1:
close #237
close #113

- Truncate long token to 8191 length, which is openai embedding context length.
- Use `get_text_embedding_batch` in llamaindex method. It is much shorter and powerful for processing batch embeddings. (It is fast!)
- Refactor duplicate id at vectordb ingest management codes.
- Refactor `sem_score` for openai embedding token limitation and batch processing.
- Rename duplicated test file name for resolving test failed.
- Delete deprecated openai embedding models. Now only support latest three embedding models.


bwook00 (approved):
LGTM! `"This is a test" * 2000` is really funny! i love that~

Eastsidegunn (approved):

PR #281, opened 2024-04-09
Add long context reorder

### LINKED ISSUE #228: 'Lost in the Middle' reorder module (or possible new node) (or possible new config option for reranker & retrieval)
[Long-context Reorder](https://python.langchain.com/docs/modules/data_connection/retrievers/long_context_reorder) from Langchain.

In the `Lost in the Middle` paper, it is well-known that reordering high-relevance context to be first and the last order in the context prompt. 

[How to?]
It can be four way.

1. Add as Reranker module. => not suggest
Its feature is simillar with Reranker because it reorder passage order.
But, in the real world scenario, it is so common to use both reranker and this 'reorder module'.

2. Add as new node => maybe? But I don't like it.
It is reordering passages, but it uses only scores from retrieval or reranking. 
It can be whole new node.
However, the cons is that it can be enforce so much complexity to choose node in config YAML file.

3. Add as 'retrieval' and 'rerank' option (or strategy)
It is possible to add as optional parameter at retrieval and reranker nodes.
It can be a new 'node parameter'
But my question is, how to test 'use lost in the middle' or not use? If we want to test this, it can be super complex at `run.py` code, because we have to test it with generator node and its metrics.

4. Add as 'prompt maker' module.
It can resolve above issue about 'evaluation' problem.
However, from now we don't have `scores` as input of prompt maker.
So we have to add `scores` as input parameter of prompt maker for making this module as prompt maker module.

I think the best option is 'prompt maker' module... What do you think?
Need to talk about it more.

bwook00:
close #228 

https://docs.llamaindex.ai/en/stable/examples/node_postprocessor/LongContextReorder/


Eastsidegunn (approved):

autorag/nodes/promptmaker/long_context_reorder.py
```diff
@@ -0,0 +1,49 @@
...
+    :param prompt: A prompt string.
+    :param queries: List of query strings.
+    :param retrieved_contents: List of retrieved contents.
+    :param retrieve_scores: List of retrieve scores.
+    :return: Prompts that made by long context reorder.
+    """
+
+    def long_context_reorder_row(_prompt: str, _query: str, _retrieved_contents: List[str],
+                                 _retrieve_scores: List[float]) -> str:
+        if not len(_retrieved_contents) == len(_retrieve_scores):
```
vkehfdl1, line 38: Just
`if len(_retrieved_contents) != len(_retrieve_scores)`

??

vkehfdl1 (approved):
Awesome. Just a little review. You don't have to fix that but LMAO

autorag/nodes/promptmaker/long_context_reorder.py
vkehfdl1: Never mind haha

vkehfdl1: Whatever, it works

PR #300, opened 2024-04-10
Add time reranker module

### LINKED ISSUE #299: Add 'Time Reranker' module
For keeping recency information, this `time_reranker` module simply reranks a retrieved passage with datetime. 
It uses a last_modified_datetime that in the corpus.parquet. 
User can specify which metadata column they want to use for reranks by time. 
It have to be an datetime object.

vkehfdl1:
close #299


bwook00 (approved):
LGTM! I can apply the same logic to the recency filter!

autorag/nodes/passagereranker/time_reranker.py
```diff
@@ -0,0 +1,21 @@
...
+from typing import List, Tuple
+
+from autorag.nodes.passagereranker.base import passage_reranker_node
+
+
+@passage_reranker_node
+def time_reranker(contents_list: List[List[str]],
+                  scores_list: List[List[float]], ids_list: List[List[str]],
+                  top_k: int, time_list: List[List[datetime]]
+                  ) -> Tuple[List[List[str]], List[List[str]], List[List[float]]]:
```
bwook00, line 11: I think we should have some documentation here!

bwook00 (changes requested):

Eastsidegunn (dismissed):

PR #319, opened 2024-04-12
Add NER PII Masking module

### LINKED ISSUE #314: Add PII-Masking module

bwook00:
close #314 

Use a Hugging Face NER model for PII Masking.


vkehfdl1 (changes requested):
Plus, are you sure NER PII Masking is 'passage filter' node? I think it is not. 
Passage filter means that 'decreasing the context' while not modifying each contexts content itself.
I think we have to make new node... like privacy filter or something...?

Plus, it is tricky to setup metric for privacy filter isn't it?

bwook00:
> Plus, are you sure NER PII Masking is 'passage filter' node? I think it is not. Passage filter means that 'decreasing the context' while not modifying each contexts content itself. I think we have to make new node... like privacy filter or something...?
> 
> Plus, it is tricky to setup metric for privacy filter isn't it?

I agree with you ... I'll give it a little more time and think about it ,,,,

autorag/nodes/passagefilter/ner_pii_masking.py
```diff
@@ -0,0 +1,41 @@
...
+
+    tasks = [mask_pii(model, contents) for contents in contents_list]
+    loop = asyncio.get_event_loop()
+    results = loop.run_until_complete(asyncio.gather(*tasks))
+    masked_contents_list = list(results)
+
+    return masked_contents_list, ids_list, scores_list
+
+
+async def mask_pii(model, contents: List[str]) -> List[str]:
```
vkehfdl1, line 32: Actually, this is useless async operation. 
I'll make a PR for resolving this kind of issue in the model rerankers...

vkehfdl1 (changes requested):

PR #334, opened 2024-04-15
Parallel processing at UPR reranker

### LINKED ISSUE #305: Parallel processing at UPR

bwook00:
close #305 

I cut to batch and realize that calculate is a single query anyway

I was trying to calculate multiple query-contents logit and found something that seemed like an easier way to do it
I tried `from multiprocessing import Pool`, what do you think?


vkehfdl1:
@bwook00 I think using `multiproessing.Pool` is the great way to enhance process speed in only *cpu* environment.
This feature looks like using multi-core cpu feature.

vkehfdl1 (changes requested):
@bwook00 I think using multiproessing.Pool is the great way to enhance process speed in only cpu environment.
This feature looks like using multi-core cpu feature.

bwook00:
> @bwook00 I think using multiproessing.Pool is the great way to enhance process speed in only cpu environment. This feature looks like using multi-core cpu feature.

How about my new method?
(use `from torch.multiprocessing import Pool, set_start_method` for gpu process)

vkehfdl1:
I used `ray` for parallel processing of CUDA model.
https://docs.ray.io/en/latest/data/examples/pytorch_resnet_batch_prediction.html
See this.

It can't process more than one query, because it effects logit probability compute score.

vkehfdl1:
Someday I can make more efficient code that using attention mask at query token.

vkehfdl1 (changes requested):
Edit docs for updated parameter

vkehfdl1 (approved):
LGTM to me 👍

PR #345, opened 2024-04-17
Move refine compressor to generator and add structured_answer_filtering parameter

### LINKED ISSUE #337: Move refine module Passage Compressor to Generator

bwook00:
close #337 

https://docs.llamaindex.ai/en/stable/examples/response_synthesizers/structured_refine/
Looking at `structured_answer_filtering` in the link above, I realized that the `Refine` module is not a module that compresses the passage, but a module that takes the passage as a base and outputs a refined answer.
-> So I decided to move `refine` to the generator Module.

As with llama index, `structured_answer_filtering` only takes a parameter to enable the function.
- I was expecting it to be the first module of the `Answer filter` Node, but when I opened it, it doesn't filter answers at all.
- It was just a function to reduce the probability of `I don't know` as the final answer.

Since `refine` puts the query and contents separately, we don't need to use the prompt created by the Prompt maker node.
- Therefore, since refine only needs an instruction prompt, I decided to take the prompt as a parameter.
- I wrote more details in the docs.


vkehfdl1:
@bwook00 I have an question. What do you mean that 'It was just a function to reduce the probability of I don't know as the final answer.' ???
How `refine` module reduce the probability of `I don't know`?? Is that what new `answer filter` node have to do? 
What is happening in the `refine` module under the hood?
If we know that, we can split `refine` module into smaller code pieces and select the best node for it.

<Why?>
I really don't like the idea that put the prompt parameter at the `generator` node. It is really confusing. Our YAML configuration is already hard to construct whole thing perfectly. So I want to avoid adding another exception. 
I belive you can find other solution to use `refine` module.

vkehfdl1 (changes requested):
I added my review at above.

PR #347, opened 2024-04-17
Create 'Passage Augmenter' node and its first module 'prev_next_augmenter'

bwook00:
close #338

- Create passage_augmenter node

- Create prev_next_augmenter
https://docs.llamaindex.ai/en/stable/examples/node_postprocessor/PrevNextPostprocessorDemo/

Plus, I found some typos in passage_filter/run.py, but they were minor so I PR'd them together.

Issue #338, opened 2024-04-16
Create 'Passage Augmenter' and its first module 'prev_next_augmenter'


bwook00:
https://docs.llamaindex.ai/en/stable/examples/node_postprocessor/PrevNextPostprocessorDemo/

PR #352, opened 2024-04-19
Add pass passage augmenter module

### LINKED ISSUE #339: Add pass augmenter module

bwook00:
close #339 

I've been merging branch #338, so please review pr #347 after it's merged.


vkehfdl1 (changes requested):
What do you think about `pass_passage_augmentor` do not perform calculating cosine similarity score at the passage augmenter node decorator?
Because it is `pass` module, I think it is better that after this module's execution, really nothing happens at all.

PR #326, opened 2024-04-13
Parallel processing at colbert reranker

bwook00:
close #308

Issue #301, opened 2024-04-11
Error when using Colbert Reranker at CUDA enabled device


vkehfdl1:
```
RuntimeError: Expected all tensors to be on the same                   
                             device, but found at least two devices, cuda:0 and cpu!                
                             (when checking argument for argument index in method                   
                             wrapper_CUDA__index_select)  
```

PR #366, opened 2024-04-23
[Hotfix] resolve colbert, flag embedding reranker batch error 

### LINKED ISSUE #364: [Hotfix] failed colbert batch one test code

### LINKED ISSUE #365: [Hotfix] failed flag_embedding batch one test code

bwook00:
close #364 
close #365 

These modules are structured to require batch to be 1 but not 2. Therefore, I tested with batch=2.
(Note that test_base_reranker had 3 example_contents instead of 2. So it was actually possible to test meaningfully with batch=2)

### DIFF
--- a/autorag/nodes/passagereranker/colbert.py
+++ b/autorag/nodes/passagereranker/colbert.py
@@ -26,11 +26,14 @@ def colbert_reranker(queries: List[str], contents_list: List[List[str]],
     :param top_k: The number of passages to be retrieved
     :param batch: The number of queries to be processed in a batch
         Default is 64.
+        The batch size must always be greater than 1.
     :param model_name: The model name for Colbert rerank.
         You can choose colbert model for reranking.
         Default is "colbert-ir/colbertv2.0".
     :return: Tuple of lists containing the reranked contents, ids, and scores
     """
+    assert batch > 1, "Batch size must be greater than 1"
+
     device = "cuda" if torch.cuda.is_available() else "cpu"
     model = AutoModel.from_pretrained(model_name).to(device)
     tokenizer = AutoTokenizer.from_pretrained(model_name)
@@ -55,14 +58,14 @@ def colbert_reranker(queries: List[str], contents_list: List[List[str]],
 
 
 def colbert_run_model(contents_list, model, tokenizer, device, batch_size: int):
-    batch_contents_list = make_batch(contents_list, batch_size)
+    flattened_queries, flattened_contents = map(list, zip(*contents_list))
+    batch_queries = make_batch(flattened_queries, batch_size)
+    batch_contents = make_batch(flattened_contents, batch_size)
     results = []
 
-    for batch_contents in batch_contents_list:
-        flattened_batch_queries, flattened_batch_contents = map(list, zip(*batch_contents))
-
+    for batch_queries, batch_contents in zip(batch_queries, batch_contents):
         # Tokenize both queries and contents together
-        feature = tokenizer(flattened_batch_queries, flattened_batch_contents, padding=True, truncation=True,
+        feature = tokenizer(batch_queries, batch_contents, padding=True, truncation=True,
                             return_tensors="pt").to(device)
 
         # Process the combined feature through the model in one go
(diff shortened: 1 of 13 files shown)


bwook00:
Add batch test code at jina, cohere, rankgpt too!

autorag/nodes/passagereranker/colbert.py
```diff
@@ -26,11 +26,14 @@ def colbert_reranker(queries: List[str], contents_list: List[List[str]],
...
     :param batch: The number of queries to be processed in a batch
         Default is 64.
+        The batch size must always be greater than 1.
     :param model_name: The model name for Colbert rerank.
         You can choose colbert model for reranking.
         Default is "colbert-ir/colbertv2.0".
     :return: Tuple of lists containing the reranked contents, ids, and scores
     """
+    assert batch > 1, "Batch size must be greater than 1"
+
```
vkehfdl1, line 36: Why the batch size must be greater than 1?

vkehfdl1 (changes requested):

autorag/nodes/passagereranker/colbert.py
bwook00: I didn't fully realize torch, so I don't know why the score is different only when the batch is 1, but it's a hotfix, so I temporarily handled it with assert like this.

Still, I was able to handle issue #301 with the flatten method above.

As I parallelize the UPR (PR #334) and learn more about torch operations, I'll fix issues that don't work in batch 1 only!

Once this PR is merged, I'll create another issue!

autorag/nodes/passagereranker/colbert.py
```diff
@@ -55,22 +58,22 @@ def colbert_reranker(queries: List[str], contents_list: List[List[str]],
...
 
         # Process the combined feature through the model in one go
         outputs = model(**feature)
         last_hidden_state = outputs.last_hidden_state
 
         # Split the embeddings back into query and content parts
-        num_queries = len(flattened_batch_queries)
+        num_queries = len(batch_queries)
         query_embedding, content_embedding = last_hidden_state.split(
             [num_queries, last_hidden_state.size(1) - num_queries], dim=1)
```
vkehfdl1, line 78: It looks like the real problem is not `batch` in here. The parallel code itself looks wrong. 

Look at #326.

You can see the original code, and the new code that you pushed.

The original code is like below.

```python
def get_colbert_score(query: str, content_list: List[str],
                      model, tokenizer) -> List[float]:
    query_encoding = tokenizer(query, return_tensors="pt")
    query_embedding = model(**query_encoding).last_hidden_state
    rerank_score_list = []

    for document_text in content_list:
        document_encoding = tokenizer(
            document_text, return_tensors="pt", truncation=True, max_length=512
        )
        document_embedding = model(**document_encoding).last_hidden_state

        sim_matrix = torch.nn.functional.cosine_similarity(
            query_embedding.unsqueeze(2), document_embedding.unsqueeze(1), dim=-1
        )

        # Take the maximum similarity for each query token (across all document tokens)
        # sim_matrix shape: [batch_size, query_length, doc_length]
        max_sim_scores, _ = torch.max(sim_matrix, dim=2)
        rerank_score_list.append(torch.mean(max_sim_scores, dim=1))

    return list(map(float, rerank_score_list))
```

The colbert model acts like 'embedding_model' in here. The `last_hidden_state` is embedding vector. 
It gets cosine similarity between each query_embedding and each content_embedding. And rerank contents by that score.

Return to your code. It is much easier you look at #326

The first thing you did wrong is tokenize query and contents at once. 

```python
# Tokenize both queries and contents together
        feature = tokenizer(flattened_batch_queries, flattened_batch_contents, padding=True, truncation=True,
                            return_tensors="pt").to(device)
```
If you decode the `feature` you can get the result like below.

```
'[CLS] what is the capital of france? [SEP] nomadamas is great team [SEP]'
```
It means, it encodes the query and content in the one sentence. And you put this feature into model 'at once'. 
I think you try to split the result of embedding at the next code, but it does not work that way at all.
Model creates embedding of the tokenized sentence. It did not create embedding each sentence separately, it embeds a whole input string. So splitting it after embedding seems useless.

In conclusion, the code itself after #326 is wrong and might get the bad results and occurs several errors.

vkehfdl1 (changes requested):

vkehfdl1 (approved):
LGTM

PR #369, opened 2024-04-24
[Hotfix] recency filter change threshold parameter type str to datetime

### LINKED ISSUE #367: [Hotfix] recency filter doesn't get threshold as string
`TypeError: strptime() argument 1 must be str, not datetime.date `

bwook00:
close #367


vkehfdl1 (approved):
LGTM

PR #395, opened 2024-04-29
Show logging at module level

### LINKED ISSUE #384: View modules progress easier
From now, user can't see the progress of each modules. 
Especially on rerankers, user think that the progress is just stucked. 
For better recognition of progression, we need to set the log for processing each modules.

vkehfdl1:
close #384

- Log for starting each module at the decorators
- delete model at UPR model
- tqdm at each reranker


bwook00 (approved):
LGTM

PR #403, opened 2024-04-30
Parallel Processing of Ko Reranker

### LINKED ISSUE #394: Parallel Processing of Ko Reranker
Ko reranker still does not support parallel processing of CUDA device.

bwook00:
close #394

### DIFF
--- a/autorag/nodes/passagereranker/koreranker.py
+++ b/autorag/nodes/passagereranker/koreranker.py
@@ -1,17 +1,19 @@
-import asyncio
 from typing import List, Tuple
 
 import numpy as np
+import pandas as pd
 import torch
+from tqdm import tqdm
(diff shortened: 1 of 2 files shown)


vkehfdl1 (changes requested):
The rerank score change when you change batch size. 
The result always be 1.0 when the batch size is 1.
The gt value is 1.0 when the batch size is 3.
The gt value is around 0.5 when the batch size is 6 or more.

This is maybe because ko-reranker calculates logits of each passages, and calculates difference between each passage with one query.

Thus, in here, you can compute various queries at once. But, the calculation of `exp_normalize` must be done one by one (per query).
Plus, the computation of exp_normalize must contain full passages from one query.

vkehfdl1 (approved):
LGTM

PR #424, opened 2024-05-10
Fix error at UPR and Colbert Reranker on GPU

### LINKED ISSUE #407: Colbert Reranker too many VRAM on H100 instance
H100 has 80gb vram. Colbert Reranker occurs OOM on the machine. It is really weird.

vkehfdl1:
close #407 

the `with torch.no_grad():` was missing.
I added it.

Plus, calculate cosine similarity after colbert embedding is pretty slow.
So, change code to use numpy for calculating cosine similarity.

I tested UPR & Colbert at H100 instance.
Plus, the colbert score did not change at all. Worked perfectly.


bwook00 (approved):
LGTM

PR #426, opened 2024-05-11
[HotFix] empty cache and return only float value at vllm.py

### LINKED ISSUE #425: [HotFix] Error at vllm.py
1. The output logprob is not float, so it can't save as parquet file
2. It does not properly empty GPU memory when destroyed.

vkehfdl1:
close #425


bwook00 (approved):
LGTM

PR #442, opened 2024-05-15
[Hotfix] Respond when the input 'scores_list' is `np.ndarray` at threshold_cutoff_pure

### LINKED ISSUE #439: [HotFix] [threshold cutoff] Error when the input 'scores_list' is np.ndarray.
In the running of AutoRAG, the input of `threshold_cutoff_pure` function's parameter `scores_list` is np.ndarray, it occurs error.

```python
def threshold_cutoff_pure(scores_list: List[float],
                          threshold: float,
                          reverse: bool = False) -> List[int]:
    if reverse:
        remain_indices = [i for i, score in enumerate(scores_list) if score <= threshold]
        default_index = scores_list.index(min(scores_list)) # error here
    else:
        remain_indices = [i for i, score in enumerate(scores_list) if score >= threshold]
        default_index = scores_list.index(max(scores_list)) # error here

    return remain_indices if remain_indices else [default_index]
```

bwook00:
close #439

### DIFF
--- a/autorag/nodes/passagefilter/threshold_cutoff.py
+++ b/autorag/nodes/passagefilter/threshold_cutoff.py
@@ -1,5 +1,7 @@
 from typing import List, Tuple
 
+import numpy as np
+
 from autorag.nodes.passagefilter.base import passage_filter_node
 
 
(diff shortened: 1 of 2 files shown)


vkehfdl1 (approved):
You can just change scores_list to python list at the first line of function...

Anyway, good 👍

PR #444, opened 2024-05-16
Resolve Korean encoding unicode error

### LINKED ISSUE #443: Problem with Korean tokenize
https://jonsyou.tistory.com/26

This happens a lot when we processing Korean documents.
Need to fix (especially on the mac)

vkehfdl1:
close #443

resolve error at casting qa dataset and corpus dataset


bwook00 (approved):
LGTM

PR #452, opened 2024-05-20
Can't use passage compressor when there is no retrieval gt values in QA datset

### LINKED ISSUE #450: bug in passage_compressor node metrics get Nan

vkehfdl1:
close #450


bwook00 (dismissed):
LGMT

bwook00 (approved):

PR #461, opened 2024-05-23
implement multi_query_expansion module

### LINKED ISSUE #227: Add 'custom multi' query expansion module
There is a [MultiQueryRetriever](https://python.langchain.com/docs/modules/data_connection/retrievers/MultiQueryRetriever) in Langchain.
It is kind of 'query expansion', which can generate several queries from given user query.

There is a default prompt, but user can write their own 'query expansion' prompt. 

From now, we only have 'hyde' and 'query_decompose' module and do not have corresponding module from `MutliQueryRetriever`.

To use this, user can adapt any LLM using query expansion method easily, just change prompt in config YAML file.

[How to Suggestion]
It can be great just slice LLM generation to line by line (using \n)
Because this module must generate more than one question, it is important to parse LLM generation.

bwook00:
close #227


vkehfdl1 (changes requested):
I think use generator modules better than use LLMPredictorType from LlamaIndex.
Because we want to change it in Hyde and question decompose anyway.

vkehfdl1 (approved):

PR #481, opened 2024-06-06
Adjust embedding batch size at YAML file

vkehfdl1:
Using `embedding_batch` at vectordb module.

It had problem at upstage embedding, and I tested using it.


bwook00 (approved):
LGTM

PR #483, opened 2024-06-08
Resolve error that tokenized tokens and logprob tokens length sometimes mismatched

vkehfdl1:
Now, tokenize the tokens from the logprob resut. Now, the length of the two (logprob and tokens) never be different.


bwook00 (approved):
LGTM

PR #501, opened 2024-06-18
feat(__init__.py): add Ollama class to the list of generator models

hongsw:
This PR helps ensure that the `run_web` command works even when using ollama.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -41,7 +41,7 @@ llama-index-embeddings-huggingface
 llama-index-llms-openai
 llama-index-llms-huggingface
 llama-index-llms-openai-like
-llama_index.llms.ollama
+llama-index-llms-ollama
 # Retriever
 llama-index-retrievers-bm25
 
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -41,6 +41,7 @@ llama-index-embeddings-huggingface
 llama-index-llms-openai
 llama-index-llms-huggingface
 llama-index-llms-openai-like
+llama_index.llms.ollama
```
vkehfdl1, line 44: Thanks @hongsw 
But, I think it should be `llama-index-llms-ollama`.
[pypi link](https://pypi.org/project/llama-index-llms-ollama/)

vkehfdl1 (changes requested):

requirements.txt
vkehfdl1: I changed this

vkehfdl1 (approved):
LGTM

vkehfdl1:
The test is passed on my local so I will merge this @bwook00

PR #513, opened 2024-06-21
window read_parquet error

Eastsidegunn:
add all pd.read_parquet param engine ='pyarrow'
actually the error is not reproduced by my window os.
but, we determine that the user's solution of the bug does not affect the entire system.
Since many Windows users experienced the same error, I modified the code.


autorag/nodes/generator/run.py
```diff
@@ -35,7 +35,7 @@ def run_generator_node(modules: List[Callable],
     node_dir = os.path.join(node_line_dir, "generator")  # node name
     if not os.path.exists(node_dir):
         os.makedirs(node_dir)
-    qa_data = pd.read_parquet(os.path.join(project_dir, "data", "qa.parquet"))
+    qa_data = pd.read_parquet(os.path.join(project_dir, "data", "qa.parquet"),engine='pyarrow')
```
bwook00, line 38: spacing 🤣

vkehfdl1: Yeah, please refactor python file before commit.

You can set this at commit tab setting at the right under corner in the pycharm.

autorag/evaluator.py
```diff
@@ -299,7 +299,7 @@ def __set_previous_result(self, node_line_names: List[str], node_names: List[Lis
 
             previous_node_dir = os.path.join(trial_path, previous_node_line, previous_node)
             best_file_pattern = f'{previous_node_dir}/best_*.parquet'
-            previous_result = pd.read_parquet(glob.glob(best_file_pattern)[0])
+            previous_result = pd.read_parquet(glob.glob(best_file_pattern)[0],engine='pyarrow')
```
vkehfdl1: space here, too

vkehfdl1:
No error on the full test on the mac

vkehfdl1 (approved):

PR #526, opened 2024-06-24
Openai token limit feature at passage augmenter and passage filter.

### LINKED ISSUE #521: Change openai embedding token limit while using openai-embed-large model

vkehfdl1:
close #521

There were no filter at passage augmenter and passage filter yet. I added openai token limit truncation at both of nodes.


bwook00 (approved):
LGTM

PR #528, opened 2024-06-24
Delete ray dependency in AutoRAG package

### LINKED ISSUE #508: Delete ray dependency at UPR reranker
ray have conflict when installing, so delete it. Because I think it is not that important.

vkehfdl1:
close #508

ray only needed in UPR reranker. Delete the ray dependency.


bwook00 (approved):
LGTM

PR #542, opened 2024-06-28
Add decorator that converts np.array and pd.Series to python list

### LINKED ISSUE #438: Decorator that convert pd.Series or np.ndarray to list recursively.
In autorag, we use python default list when data processing.
When we use pandas or numpy, it is really easy that lists convert to np.array, tuple, or pd.Series, and it occurs unexpect error easily.

Develop decorator that converts all Iterable types to list, even the 2-d or 3-d list recursively. 
And using it on the functions for preventing that kind of errors.

vkehfdl1:
close #438


bwook00 (approved):
LGTM

Issue #418, opened 2024-05-03
Resolve errors with BLEU Score values greater than 1


bwook00:

Eastsidegunn:
in sacrebleu precision calculation, multiply it by 100.

Eastsidegunn:
The currently implemented bleu will be renamed to sacrebleu and another bleu will be implemented.

vkehfdl1:
Implemented. BLEU is now computed via sacrebleu (`from sacrebleu.metrics.bleu import BLEU`) with a configurable `tokenize` parameter, separated in commit 8da9c26a ("separate bleu and sacrebleu") and exposed upstream via #549. Closing as completed.

PR #549, opened 2024-06-29
make sacrebleu parameter available in upper stream

Eastsidegunn:


autorag/evaluation/metric/generation.py
```diff
@@ -64,12 +64,25 @@ def compute_score(gt: List[str], pred: str) -> float:
     return result
 
 
-@generation_metric
-def bleu(gt: List[str], pred: str, **kwargs) -> float:
+#@generation_metric
```
vkehfdl1, line 67: delete annotation of generation_metric decorator.
And please use `convert_inputs_to_list` decorator at here. It needs to be a python list for reducing error.

autorag/evaluation/metric/generation.py
```diff
@@ -64,12 +64,25 @@ def compute_score(gt: List[str], pred: str) -> float:
...
+
+    :param tokenize: The tokenizer to use. If None, defaults to language-specific tokenizers with '13a' as the fallback default. check #https://github.com/mjpost/sacrebleu/blob/master/sacrebleu/metrics/bleu.py
+    :param smooth_method: The smoothing method to use ('floor', 'add-k', 'exp' or 'none').
+    :param smooth_value: The smoothing value for `floor` and `add-k` methods. `None` falls back to default value.
+    :param max_ngram_order: If given, it overrides the maximum n-gram order (default: 4) when computing precisions.
+    :param trg_lang: An optional language code to raise potential tokenizer warnings.
+    :param gt: gt is reference. A sequence of reference documents with document being
+    defined as a sequence of reference strings. If given, the reference n-grams
+    and lengths will be pre-computed and cached for faster BLEU computation
+    across many systems.
```
vkehfdl1, line 80: It looks like there are no `gt` parameter at this function

autorag/evaluation/metric/generation.py
```diff
@@ -64,12 +64,25 @@ def compute_score(gt: List[str], pred: str) -> float:
...
 
-@generation_metric
-def bleu(gt: List[str], pred: str, **kwargs) -> float:
+#@generation_metric
+def bleu(generation_gt: List[List[str]], generations: [str], tokenize: str|None = None, smooth_method: str = 'exp', smooth_value: Optional[float] = None, max_ngram_order: int = 4, trg_lang: str = '', **kwargs) -> List[float]:
     """
-    Compute bleu score for generation.
+    Computes the BLEU metric given pred and ground-truth.
+
+    :param tokenize: The tokenizer to use. If None, defaults to language-specific tokenizers with '13a' as the fallback default. check #https://github.com/mjpost/sacrebleu/blob/master/sacrebleu/metrics/bleu.py
```
vkehfdl1, line 72: Please add annotation for generation_gt and generations.
```
:param generation_gt: A list of ground truth.
        Must be 2-d list of string.
        Because it can be a multiple ground truth.
:param generations: A list of generations that LLM generated.
```

vkehfdl1 (changes requested):
Plus, add some test case that using BLEU new options

vkehfdl1 (approved):
LGTM
But we still have this problem.
image

Need to fix at #418

PR #574, opened 2024-07-15
Refactoring hybrid retrieval with more advanced optimizations

### LINKED ISSUE #557: Redesign hybrid-cc related retrievalmodule
The current hybrid module have some problems. It is all hybrid cc (convex combination) related modules except RRF. 
I think we need to change and re-design a whole module. 
Here is my proposed design.

=> [related paper](https://arxiv.org/pdf/2210.11934)

## 1. Make hybrid_cc module to one and get an 'normalize' parameter.

All hybrid_cc related module have different only at the normalization process. If you select different normalization method, the result will be slightly different. Keep this as parameter. 

We must support below normalization method.

- mm : Min-Max normalization
- tmm : Theoretical min-max normalization
- z : z-score normalization
- dbsf : 3-sigma normalization

## 2. Test **many** weights at once

We get semantic and lexical scores from previous run. So, fusion takes so little time to compute. But, fusion quality differs so much , up to 0.2 in NDCG.
So, it if crucial to do experiment on the many weights. 
It will be great we set the weights quantity and do experiment on that. 
Like, setting it as 100, we will experiment each weights from 0.0 to 1.0 a hundred times. 
image

The actual weight optimization will be in the module execution. As result, you will only see the result of the optimization. Remain as you set the weight hyperparameter. Let me think how to save the great weight parameter at the process.

## 3. Specify which module is lexical and which module is semantic.

We have to specify it since we will implement tmm normalization. 
We have to test the name of the retrieval module, and set the theoretical minimum. 
This way, it will much clearer.

vkehfdl1:
close #557

1. Delete hybrid dbsf and hybrid rst => each method integrate to hybrid cc as different normalize methods
2. Optimize weight with input range at the retrieval/run.py. Can optimize both hybrid_rrf and hybrid_cc
3. Running optimized hybrid retrieval at extracted YAML file and Runner.run.
4. Do not use `target_module_params` and auto-detect lexical and semantic modules to fuse.
5. Fix YAML file for newer version.
6. Fix documentation for newer hybrid retrieval.

=> After review, this will be new feature at newer version 0.2.10


bwook00 (approved):
Great LGTM

PR #580, opened 2024-07-19
Impute relevance score when there is no score at the hybrid fusion.

### LINKED ISSUE #573: Get relevance score of all passages when execute hybrid retrieval
When we run hybrid retrieval, the non-duplicated passage that do not have relevance score turns to be zero.
But that is not precise way to calculate hybrid retrieval score.

So, getting relevance score when the passage is not the top-k passage is essential for more precise hybrid retrieval.

vkehfdl1:
close #573


bwook00 (approved):
LGTM

PR #596, opened 2024-08-04
Add few description for better understanding of AutoRAG

### LINKED ISSUE #560: Refactor documentation for better understanding
Thanks for Filipp who leave kind comment 

---

Filipp Trigub — 2024.07.05. 오전 1:32
Hi @Jeffrey Kim, thanks for the quick response.

In a nutshell I would like a more complete docu. As a dev, I understand that this is not high priority at all (and annoying). 

As a user however, I am truly perplex, why this project is not much more widely known. Difficulties in setting it up and running it might be one explanation. Never underestimate the laziness of tech people 😄 

Some small examples I've encountered:
the prompt maker node https://docs.auto-rag.com/nodes/prompt_maker/prompt_maker.html lists generation modules as a setting. I believe it is optional, but that is not clear at all. Would wish for clear distinction between optional and mandatory settings.
in https://docs.auto-rag.com/data_creation/data_format.html it is not clearly indicated that the doc_id in corpus must be one of the ids in retriaval_gt. Is that not obvious? On it's own, probably, but when you are already dealing with all the other info, it might get lost. Happened to me.
in https://docs.auto-rag.com/optimization/optimization.html is is not clear, which nodes are mandatory and which ones are not (or maybe they all are and I am mistaken?).

Additionally, but not of any importance:
link at llama_index is outdated here https://docs.auto-rag.com/nodes/generator/llama_index_llm.html and here https://docs.auto-rag.com/local_model.html#supporting-llm-models

Hope that helps. I really think, people should use AutoRAG much more and the good thing is that it is actually easy to use, once you've understood the nuances! 

---

Let's fix this!

vkehfdl1:
close #560

### DIFF
--- a/docs/source/data_creation/data_format.md
+++ b/docs/source/data_creation/data_format.md
@@ -76,6 +76,14 @@ it is okay to save it as 1-d list or just string.
 If you save it as 1-d list, it treats as 'and' operation.
 ```
 
+```{attention}
+The ids of retrieval_gt must be included in the corpus dataset as `doc_id`.
+
+When AutoRAG starts to evaluate, it checks the existence of the retrieval ids in the corpus dataset.
+
+You MUST match the retrieval_gt ids with the corpus dataset.
+```
+
 This column is crucial because AutoRAG evaluate retrieval performance with this column.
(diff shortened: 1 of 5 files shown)


bwook00 (approved):
LGTM

PR #607, opened 2024-08-12
Add New feature: existing_query_df and local parameters

skmanzg:
I know you said it should be the form of an independent function, but I have been focusing on adding new features while preserving existing styles instead of creating another new function. Not to mention there will be many unnecessary duplicated code lines if I did that independently. So, I would like you to review it first and then discuss it later.😅

### DIFF
--- a/autorag/data/qacreation/llama_index.py
+++ b/autorag/data/qacreation/llama_index.py
@@ -4,6 +4,7 @@ import random
 from typing import Optional, List, Dict, Any
 
 import pandas as pd
+from llama_index.core.base.llms.types import ChatMessage, MessageRole
 from llama_index.core.service_context_elements.llm_predictor import LLMPredictorType
 
 from autorag.utils.util import process_batch
@@ -53,37 +54,28 @@ def generate_qa_llama_index(
     return results
 
 
-
-def generate_qa_with_existing_queries(
+def generate_answers(
     llm: LLMPredictorType,
     contents: List[str],
     queries: List[str],
-    prompt: Optional[str] = None,
-    max_retries: int = 3,
     batch: int = 4,
 ) -> List[List[Dict]]:
     """
-    Generate a qa set from the list of contents using existing queries.
+    Generate qa sets from the list of contents using existing queries.
 
     :param llm: Llama index model
     :param contents: List of content strings.
     :param queries: List of existing queries.
-    :param prompt: The prompt to use for the qa generation.
-    :param max_retries: The maximum number of retries when generated question number is not equal to the target number.
     :param batch: The batch size to process asynchronously.
     :return: 2-d list of dictionaries containing the query and generation_gt.
     """
-    if prompt is None:
-        prompt = open(os.path.join(package_dir, "llama_index_default_prompt.txt"), 'r').read()
 
     tasks = [
-        async_qa_gen_llama_index_with_query(content, llm, prompt, max_retries, query)
+        generate_basic_answer(llm, content, query)
         for content, query in zip(contents, queries)
     ]
-    
     loops = asyncio.get_event_loop()
     results = loops.run_until_complete(process_batch(tasks, batch))
-    
     return results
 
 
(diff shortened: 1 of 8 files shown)


skmanzg:
bug founded: openAI shape ERROR, cuda 1 setting. (will fix soon)

autorag/data/qacreation/llama_index.py
```diff
@@ -8,6 +9,9 @@
 
 from autorag.utils.util import process_batch
 
+
+nest_asyncio.apply()
```
vkehfdl1, line 13: Recommend to delete this if you add this for jupyter notebook.
Unwanted nest asyncio at non-jupyter environment.

autorag/data/qacreation/llama_index.py
```diff
@@ -18,6 +22,7 @@ def generate_qa_llama_index(
         question_num_per_content: int = 1,
         max_retries: int = 3,
         batch: int = 4,
+        queries: Optional[List[str]] = None,
```
vkehfdl1, line 25: It is better to separate new function when queries is not None.

vkehfdl1 (changes requested):
1. It will be great you separate this feature to independent function
2. Recommend use chromadb.Client instead of FAISS. 
AutoRAG didn't have FAISS dependency. And for embedding search, it looks better to use chromadb like any other AutoRAG functions.

vkehfdl1:
Plus, you need to make corresponding test code at `tests` folder!

skmanzg:
So, I followed most of your requirements:
1. New feature should be independent✔️
2. New feature use Chroma instead of FAISS✔️
3. add test code....? 🤔 (not... yet)


This is a sample code in the JupyterNotebook below:

```
from autorag.data.qacreation import generate_qa_with_existing_queries, make_qa_with_existing_queries
from llama_index.llms.vllm import Vllm
from llama_index.llms.openai import OpenAI
from dotenv import load_dotenv
import pandas as pd

# if you use jupyter, you must use this.
import nest_asyncio
nest_asyncio.apply()


# Prepare your own dataset! (Eaxamples below)
# corpus_df = pd.read_parquet('corpus.parquet')
# df = pd.read_parquet('q.parquet')



prompt = """
Answer the questions based on the contents.

Questions:
{{query}}

contents:
{{text}}

Response:
"""

# if you use OpenAI, you need this.
# load_dotenv()
# llm = OpenAI(model='gpt-3.5-turbo')

# local model or other model other than OpenAI
llm = Vllm(
    model="your model path",
    # tensor_parallel_size=4,    # this option is not available for all. if it is not work, delete it.
    temperature=0.7,
    max_new_tokens=256,
    vllm_kwargs={
        "swap_space": 1,
        "gpu_memory_utilization": 0.8,
    },
)

qa_df = make_qa_with_existing_queries(
    corpus_df, 
    existing_query_df=df, 
    content_size=100, 
    qa_creation_func=generate_qa_with_existing_queries,
    llm=llm, 
    prompt=prompt,
    local=True,
    top_k=5,
    output_filepath= "qa.parquet",
)
```

I haven't tested till the end in OpenAI envrionment due to a lack of tokens. There might be bugs.

vkehfdl1:
@skmanzg I'll check it and add test case + documentation. Thanks

skmanzg:
> @vkehfdl1 I'll check it and add test case + documentation. Thanks

I've read what you've done. I understand you fix the most to fit in AutoRAG system. but is there a part for saving embedding vector DB from Corpus df? In my case, I often had to test or change some parameters with the same data. Since creating corpus vector DB can takes a lot if its size is big, I think there should be save and load Corpus vector DB system or at least, optionally. I might misunderstand what you've done because I do not know well about overall AutoRAG function structures anyway! ... and if that is the case, could you explain to me how it can save and load? 

Thank you in advance.

vkehfdl1:
@skmanzg 
I agree.
If so, I'll make a chromadb Collection as parameter. 
If you want to use it locally, or temp collection, or remote server Chroma, the whole case will be cover if we get the collection.
(Users have to write their own code that loading chromadb collection, but it is piece of cake)

vkehfdl1:
@bwook00 Review please ^^

skmanzg:
> (Users have to write their own code that loading chromadb collection, but it is piece of cake)
I doubt about it.


@vkehfdl1 so.. I see collection is for loading DB. but there is no way to save DB from this function... right?

vkehfdl1:
All test was passed at the local.
This is made by outside contributor, so cause error at unit test (No OpenAI API Key)

vkehfdl1 (approved):
@skmanzg Nothing changed on your force-push right?

skmanzg:
> @vkehfdl1 Nothing changed on your force-push right?

Not at all. I just updated the branch with rebase.

vkehfdl1:
@skmanzg I am bypassing auto-testing (due to the OPENAI API KEY) and merge it.

Issue #613, opened 2024-08-13
[Data Creation Refactoring] Re-design QA generation process


vkehfdl1:
Current library is little bit complicated if you want to make a custom QA generation process. 
Making custom QA creation process must be easy. 

"The process will change a lot.”

I think the legacy codes will be deprecated and delete at AutoRAG 0.3.0.

- [x] #661
- [x] #662
- [x] #673
- [x] #674
- [x] #675
- [x] #676
- [x] #677 
- [x] #683

vkehfdl1:
It can be a kind of monad.
There is a dataclass. Let's name it.... `Dataset` 
`Dataset` contains `qa` and `corpus`.
Every function in `Dataset` returns `Dataset` instance, but modified `Dataset` instance. 
Every function must be `pure function`. Its input is `Dataset` instance, and output is modified `Dataset` instance.
We can load and chunk from the class function.
We can make question, make answers. 
After that, you can convert it to qa_df and corpus_df (dataframes for AutoRAG)

For example (concept)

```python
Dataset.from_directory('path/to/dir').recursive_chunk(chunk_size=512, overlap=128).sample_corpus(n=500).create_factoid_question(n=50).create_basic_answer().save_to_parquet(qa_data_path='./data/qa.parquet', corpus_data_path='./data/corpus.parquet')
```

vkehfdl1:
- [ ] Add basic schema and add generate answers
- [ ] Add generate qa set.

vkehfdl1:
## The basic workflow

1. Parsing the raw document and make `Raw`
2. Initial chunk the corpus from `Raw`
3. Sample some chunks to generate questions
4. Generate questions from the selected chunks
5. Generate answer and finish the initial `QA` set
6. Add more `Corpus` to use different chunking methods
7. Map new retrieval_gt from the new `Corpus` at `QA`

Looks like this below.


```python
openai_client = AsyncOpenAI()
parsing_result = Raw(pd.read_parquet('./parse.parquet'))
initial_corpus = parsing_result.chunk(lambda data: recursive_split(data, chunk_size=128, chunk_overlap=24))
initial_qa = initial_corpus.sample(
    lambda data: random_single_hop(data, n=50)
).batch_apply(
    lambda row: factoid_query_gen(row, openai_client, lang='ko')
).batch_apply(
    lambda row: lambda row: make_concise_gen_gt(row, openai_client, lang='ko')
)

[...]

PR #620, opened 2024-08-16
Add optional parameter 'exist_gen_gt' at make_qa_with_existing_queries function.

### LINKED ISSUE #618: Add optional parameter 'exist_generation_gt' at `make_qa_with_existing_queries` function

bwook00:
close #618 

- change function name `make_qa_with_existing_queries` to `make_qa_with_existing_qa`
- add optional parameter exist_gen_gt
- add test code with gen_gt
- update docs

### DIFF
--- a/tests/autorag/data/qacreation/test_base_qacreation.py
+++ b/tests/autorag/data/qacreation/test_base_qacreation.py
@@ -10,7 +10,7 @@ import pandas as pd
 import pytest
 from llama_index.core.base.llms.types import CompletionResponse
 
-from autorag.data.qacreation import make_single_content_qa, generate_qa_llama_index, make_qa_with_existing_queries, \
+from autorag.data.qacreation import make_single_content_qa, generate_qa_llama_index, make_qa_with_existing_qa, \
     generate_answers
 from autorag.utils import validate_qa_dataset
 from tests.mock import MockLLM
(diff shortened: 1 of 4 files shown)


vkehfdl1 (changes requested):
In this stage, I think we need to refactor `make_qa_with_existing_queries` again.
Actually, it is better to make functions that just do one thing, right?
However, the `make_qa_with_existing_queries` do **two** things. 
1. Make pseudo retrieval gt with Chroma
2. Generate answer with LLM (when there is no generation gt)

The reason why this function gets complicated and complex is that this function supposed to be designed to do two things at one function. 
So, I suggest we must divide this two. 

(The issue itself, which considering situation that there can be existing generation gt is great and totally reasonable)

First is making retrieval gt
And second is making generation gt.

And anyway, I think we have to refactor whole data generation process at #613
Current data generation process is way more complex + hard to customize. Let's design much better one.

@bwook00 So, if you don't want to work on this PR and just do it at #613, it's up to you.

tests/autorag/data/qacreation/test_base_qacreation.py
```diff
@@ -102,11 +102,24 @@ def test_make_qa_with_existing_queries(qa_parquet_filepath):
...
+    corpus_df = pd.read_parquet(os.path.join(resource_dir, "corpus_data_sample.parquet"), engine='pyarrow')
+    query_df = pd.read_parquet(os.path.join(resource_dir, "qa_data_sample.parquet"), engine='pyarrow')
+    qa_df = make_qa_with_existing_qa(
+        corpus_df, query_df, content_size=5, exist_gen_gt=True,
+        output_filepath=qa_parquet_filepath, upsert=True,
+    )
+    validate_qa_dataset(qa_df)
+    assert len(qa_df) == 5
+    assert all(len(elem) == 3 for elem in qa_df['retrieval_gt'].apply(lambda x: x[0]).tolist())
+    assert all((elem in query_df['query'].tolist()) for elem in qa_df['query'].tolist())
```
vkehfdl1, line 115: I think you need to add test case for testing existing generation_gt is properly merged at output qa_df

vkehfdl1 (changes requested):

vkehfdl1 (approved):

Issue #621, opened 2024-08-19
[Feature Request] Add pylint or styling guide for consistent code format (with contributors)


vkehfdl1:
**Is your feature request related to a problem? Please describe.**
The maintainers have to reformat again when there are some contributors because we use Pycharm code reformatting. 

**Describe the solution you'd like**
Consider to use [ruff](https://github.com/astral-sh/ruff).
Super fast rust-based python linter. 

**Describe alternatives you've considered**
We can use `black` as well.

PR #626, opened 2024-08-20
Add running web interface in Runner with Gradio

### LINKED ISSUE #616: How can I add custom embedding model when running web
I have custom huggingface embedding model in my pipeline.yaml.  
When running as api server, I can add my custom model like below
```python
autorag.embedding_models["custom model"] = LazyInit(
    HuggingFaceEmbedding, model_name="custom_model"
)
runner = Runner.from_yaml("./6/pipeline.yaml")
runner.run_api_server()
```
But run_web support cli only, so I can't add custom model like above.  
How can I add custom model?
Is there any reason why `run_web` is not supported by the Python SDK, like `run_api_server()`?

rjwharry:
close #616   
  
There are limitations to running the web interface with Streamlit when using custom models (embedding, llm).
1. Use Gradio
This PR uses Gradio as the web interface because it can be run in the same process that defined the custom model.
1. Fix error of `get_event_loop`
When running web with Gradio, I faced an error in `loop = asyncio.get_event_loop()` in the autorag module. Here is the detailed error log.
```bash
loop = asyncio.get_event_loop()
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/anaconda3/envs/autorag/lib/python3.11/asyncio/events.py", line 681, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
RuntimeError: There is no current event loop in thread 'AnyIO worker thread'.
```
I've modified it to create a new event loop when there is no running one, since it presumably doesn't have a running one, and the [official asyncio documentation recommends](https://docs.python.org/3/library/asyncio-eventloop.html) using `get_running_event_loop()` rather than `get_event_loop()`.  
1. Apply ruff linter
I applied ruff linter mentioned in #621.

### DIFF
--- a/autorag/__init__.py
+++ b/autorag/__init__.py
@@ -14,64 +14,73 @@ from llama_index.llms.openai import OpenAI
 from llama_index.llms.openai_like import OpenAILike
 from rich.logging import RichHandler
 
-version_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'VERSION')
+version_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "VERSION")
 
-with open(version_path, 'r') as f:
-    __version__ = f.read().strip()
+with open(version_path, "r") as f:
+	__version__ = f.read().strip()
 
 
 class LazyInit:
-    def __init__(self, factory, *args, **kwargs):
-        self._factory = factory
-        self._args = args
-        self._kwargs = kwargs
-        self._instance = None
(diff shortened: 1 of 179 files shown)


vkehfdl1:
Hi @rjwharry 
Maybe there is so much conflict after I merged Linter to main.
It will be better to close the PR and make a new one from current main branch. 
It looks unsolvable manually....

rjwharry:
Okay, I close this PR by myself and recreate new one

Issue #636, opened 2024-08-22
Can I use certain tags from the ollama model in the generator node? (generator node에서 ollama 모델의 특정 태그를 사용 할 수 있나요?)


khlee369:
I want to use a specific ```model:tag``` provided by ollama.

Ref
- https://ollama.com/library/qwen2:72b

However, if you use the ollama specific tag as shown below, i got an error.
```
      modules:
        - module_type: llama_index_llm
          llm: ollama
          model: [llama3, qwen2:72b]
          temperature: 0.7
          batch: 1
```

<details>
<summary>error message</summary>

```
[08/22/24 12:35:30] ERROR    [__init__.py:73] >> Unexpected exception                                                         __init__.py:73
                             ╭───────────────────────────── Traceback (most recent call last) 

.... 
                             ReadTimeout                                                                                                    
                                                                                                                                            
                             The above exception was the direct cause of the following exception:                                           
                                                                                                                                            
                             ╭───────────────────────────── Traceback (most recent call last) ──────────────────────────────╮               
                             │ /root/workspace/corporate-llm/autorag/main.py:27 in <module>                                 │               
                             │                                                                                              │               
                             │   24                                                                                         │               
                             │   25                                                                                         │               
                             │   26 if __name__ == '__main__':                                                              │               
                             │ ❱ 27 │   main()                                                                              │               
                             │   28                                                                                         │               
                             │                                                                                              │               
                             │ /opt/conda/envs/llm_hugg/lib/python3.10/site-packages/click/core.py:1157 in __call__         │               
                             │                                                                                              │               
                             │   1154 │                                                                                     │               
                             │   1155 │   def __call__(self, *args: t.Any, **kwargs: t.Any) -> t.Any:                       │               
                             │   1156 │   │   """Alias for :meth:`main`."""                                                 │               
                             │ ❱ 1157 │   │   return self.main(*args, **kwargs)                                             │               
                             │   1158                                                                                       │               
                             │   1159                                                                                       │               
                             │   1160 class Command(BaseCommand):                                                           │               
                             │                                                                                              │               
                             │ /opt/conda/envs/llm_hugg/lib/python3.10/site-packages/click/core.py:1078 in main             │               
                             │                                                                                              │               
                             │   1075 │   │   try:                                                                          │               
                             │   1076 │   │   │   try:                                                                      │               
                             │   1077 │   │   │   │   with self.make_context(prog_name, args, **extra) as ctx:              │               
[...]

PR #638, opened 2024-08-23
fix ```Ollama``` default request_timeout from 30 to 3000

khlee369:
Relevant Issue
- #636 

It seems that the issue is related to the default setting of ```DEFAULT_REQUEST_TIMEOUT = 30.0``` in ```llama_index.llms.ollama```. When loading a model with [ollama](https://ollama.com/) and performing inference, if the process takes longer than 30 seconds, the coroutine fails to await and the request is immediately terminated.

To temporarily resolve this issue, you can either modify ```DEFAULT_REQUEST_TIMEOUT = 30.0``` in ```llama_index.llms.ollama.base``` (e.g., change it to 3000.0), or set the request_timeout in the config.yaml file under the generator node.

```
      modules:
        - module_type: llama_index_llm
          llm: ollama
          model: qwen2:72b
          temperature: 0.7
          request_timeout: 3000
          batch: 1
```

This PR temporarily changes the default ```request_timeout``` of the ```Ollama``` class to 3000 seconds.

### DIFF
--- a/autorag/__init__.py
+++ b/autorag/__init__.py
@@ -63,11 +63,15 @@ embedding_models = {
 	"huggingface_bge_m3": LazyInit(HuggingFaceEmbedding, model_name="BAAI/bge-m3"),
 }
 
+class CustomOllama(Ollama):
+    def __init__(self, **kwargs):
+        super().__init__(request_timeout=3000, **kwargs)
+
 generator_models = {
 	"openai": OpenAI,
 	"huggingfacellm": HuggingFaceLLM,
 	"openailike": OpenAILike,
-	"ollama": Ollama,
+	"ollama": CustomOllama,
 	"mock": MockLLM,
 }
 


khlee369:
In my view, this is a very urgent issue. Not only have the operations of the Ollama tag been affected, but suddenly the inference of all Ollama models has been interrupted, leading me to reconsider using AutoRAG. This PR does not fully resolve the problem, but it provides a temporary fix.

It's fine to close this PR later after implementing a better solution.

vkehfdl1:
@khlee369 
The code you provided will work. 
But, is it really great choice to increase request timeout to 3000 seconds? Since the LlamaIndex default is 30 second.
I think the better way is to update documentation to prevent similar error. Since you can set request_timeout at config YAML file, I think we don't have to increase it dramatically.

So, in my think, as temporal fix, just update document and leave solution at the issue #636. 

Then, we can find the better solution for this.

khlee369:
@vkehfdl1 

I agree with you. This is only a temporary workaround and not an ideal solution.

When I first encountered this error, I searched the [AutoRAG documentation](https://docs.auto-rag.com/nodes/generator/generator.html) but was unable to find the relevant information. As you suggested, it seems better to explicitly state that increasing the request_time can resolve the issue rather than implicitly extending it.

If you have the time in the future, it might be worth considering an update to the [AutoRAG documentation](https://docs.auto-rag.com/nodes/generator/generator.html).

I will include the above solution (changing config.yaml) in the issue, as per your recommendation.

vkehfdl1:
I made an issue for the documentation #645

PR #652, opened 2024-08-26
fix an issue with loading HuggingfaceLLM models

jis478:
## Issue
When using a model from HuggingFace, the repo ID is not correctly recognized by AutoRAG, so the model fails to load.

## Cause analysis
- `HuggingFaceLLM` class from `llama_index` requires either `model` or `model_name`, and the repo ID (a string) must be passed to HuggingFaceLLM via `model_name`, not `model`.
- The current code fails to load the model because it uses `model` as the repo ID (it works okay for openai as it uses `model` for the same purpose.)

## Fix
- Add exception like following:
```
if llm == "huggingfacellm":
    kwargs["model_name"] = kwargs.pop("model")
    kwargs["tokenizer_name"] = kwargs["model_name"]
```


vkehfdl1:
@jis478 
Looks like we can just make YAML file with `model_name` parameter, isn't it?
Like below

```yaml
      modules:
        - module_type: llama_index_llm
          llm: huggingfacellm
          model_name: mistralai/Mistral-7B-Instruct-v0.2
```

Plus, if you want to test both openai or ollama with llama_index_llm and huggingface llm, you can just make an YAML file like below.

```yaml
      modules:
        - module_type: llama_index_llm
          llm: huggingfacellm
          model_name: mistralai/Mistral-7B-Instruct-v0.2
        - module_type: llama_index_llm
          llm: ollama
          model: [qwen:4b, gemma:2b]
```

jis478:
> @jis478 Looks like we can just make YAML file with `model_name` parameter, isn't it? Like below
> 
> ```yaml
>       modules:
>         - module_type: llama_index_llm
>           llm: huggingfacellm
>           model_name: mistralai/Mistral-7B-Instruct-v0.2
> ```
> 
> Plus, if you want to test both openai or ollama with llama_index_llm and huggingface llm, you can just make an YAML file like below.
> 
> ```yaml
>       modules:
>         - module_type: llama_index_llm
>           llm: huggingfacellm
>           model_name: mistralai/Mistral-7B-Instruct-v0.2
>         - module_type: llama_index_llm
>           llm: ollama
>           model: [qwen:4b, gemma:2b]
> ```

That's one option, but I think allowing both ('model' and 'model_name') in the YAML file might cause confusion as they serve the same purpose?

vkehfdl1:
@jis478 
Yes I agree with that using both `model` and `model_name` can be confusing.
However, is there any reason that `LlamaIndex` using different parameter name at first place? 
Since it is LlamaIndex LLM module, it can be more confusing if we allowing use `model` at `HuggingFaceLLM` by our own.

jis478:
@vkehfdl1 

I've just checked that interestingly the `openAI` class in `LlamaIndex`  does not use `model_name` at all but use 'model` as a string, so it seems that 'model' is used interchangeably.

With this PR we are not allowing to use 'model' at HuggingFaceLLM by our own, but we are actually fixing the current inconsistency with `LlamaIndex`. The `HuggingFaceLLM` class in `LlamaIndex` expects `model_name` to be a string and `model` to be a loaded HuggingFace model (`model` is None for AutoRAG in this case and 'LlamaIndex' safely allows 'model' to be None in the code.)

I believe it is better for AutoRAG to keep the current naming conventions in the YAML file for the sake of clarity and pass the proper string `model_name` onto 'HuggingfaceLLM` in `generator`.

jis478:
@vkehfdl1 any thoughts on this issue?

autorag/nodes/generator/base.py
```diff
@@ -41,6 +41,9 @@ def wrapper(
 					"You can check valid llm names from autorag.generator_models."
 				)
 			batch = kwargs.pop("batch", 16)
+			if llm == "huggingfacellm":
+				kwargs["model_name"] = kwargs.pop("model")
```
vkehfdl1, line 45: What happened the user use both `model` and `model_name`? Then what is become first if two are different.

autorag/nodes/generator/base.py
```diff
@@ -41,6 +41,9 @@ def wrapper(
...
+				kwargs["tokenizer_name"] = kwargs["model_name"]
```
vkehfdl1, line 46: I don't know why you set input model_name to tokenizer_name, since there is parameter `tokenizer_name` at LlamaIndex `HuggingfaceLLM`

vkehfdl1 (changes requested):
I searched about why they don't use `model` and it turns out at LlamaIndex they can put their own model instance instead of just model_name.
And we can't load model instance at YAML file, so the feature makes sense.

It took some time because I asked at LlamaIndex discord community.

autorag/nodes/generator/base.py
vkehfdl1: Now understand why. Agree

vkehfdl1 (approved):

bwook00 (approved):
LGTM! Thanks for your contribute 👍

PR #657, opened 2024-08-28
Replace FastAPI with Flask

### LINKED ISSUE #640: [Feature Request] Delete nest_asyncio dependency at library (FastAPI server execution)
Using `nest_asyncio` actually breaks the design of the `asyncio`. But we had to use it because of the FastAPI server.
In this issue, we have to figure out how to use FastAPI server and AutoRAG without using `nest_asyncio` since it is not intended usage of `asyncio` [Relevant Link](https://stackoverflow.com/questions/59740704/correct-use-constraints-of-use-of-nest-asyncio)

We can just use it. Actually this issue is not that serious yet.

**Additional context**

rjwharry:
close #640 

- FastAPI makes it impossible to avoid asyncio event loop crashes without nest_asyncio.  
- To delete nest_asyncio dependency, I replace FastAPI with Flask.  
- Flask doesn't run the event loop itself, which fundamentally solves this issue.

### DIFF
--- a/autorag/deploy.py
+++ b/autorag/deploy.py
@@ -221,7 +221,7 @@ class Runner:
 		.. Code:: json
 
 		    {
-		        "Query": "your query",
+		        "query": "your query",
 		        "result_column": "generated_texts"
 		    }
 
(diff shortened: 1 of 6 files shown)


vkehfdl1 (approved):
LGTM
It works perfectly
Thanks :)

vkehfdl1:
I merged it. It passed whole test in my local.

PR #668, opened 2024-09-02
Add table hybrid parse module

bwook00:
Plan to write docs after code review ^^


autorag/data/parse/table_hybrid_parse.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def get_each_module_result(
+	module: str, module_params: Dict, data_path_glob: str
+) -> Tuple[List[str], List[str]]:
+	module_params["module_type"] = module
+
+	data_path_list = glob(data_path_glob)
+	if not data_path_list:
+		raise FileNotFoundError(f"data does not exits in {data_path_glob}")
```
vkehfdl1, line 93: What if there is no page that has table.
Or there is no page without table.
Then, one of the directory between `text_dir` and `table_dir` will be empty and there will be an `FileNotFoundError` at here.
But it looks not good for me since it is intended feature. 

I think in this case, just return the empty list as a result.

bwook00: That's a very good idea!

autorag/data/parse/table_hybrid_parse.py
```diff
@@ -0,0 +1,100 @@
...
+		)
+
+		# Extract table pages
+		text_results, text_file_names = get_each_module_result(
+			text_parse_module, text_params, os.path.join(text_dir, "*")
+		)
+
+		# Merge parsing results of PDFs with and without tables
+		texts = table_results + text_results
+		file_names = table_file_names + text_file_names
```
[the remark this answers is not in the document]
bwook00: Even though the file_names are in the temp directory, they go through `file_name = os.path.basename(data_path)`, so I think it's okay because it only stores the .pdf filename.

For example, if I have a 3-page `Newjeans_book.pdf`, 
file_names will only contain ['Newjeans_book_page_1.pdf`, 'Newjeans_book_page_2.pdf`, 'Newjeans_book_page_3.pdf`] !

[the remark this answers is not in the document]
vkehfdl1: I think it is tricky to trace back the original pdf when the hybrid parser returns that.

1. The parser must have similar input and output schema like another parsers. Agree?
2. Let's imagine we want to make a GUI visualizer that want to look parsing result easily. We have to load each parsing result, and also the origin file. 
If the file name looks like this 'havertz_0_1.pdf' then the hybrid parser filename result looks like this. 'havertz_0_1_1.pdf' for first page.
Then how can you know the pdf original name is 'havertz_0_1.pdf' or 'havertz_0_1_1.pdf'??
Plus, even worse, what if there is another pdf that has different meaning that has name 'havertz_0_1_1.pdf' alerady?
3. So, in my opinion, the output information of where is the file comes from **must be** traceable. 
4. The easiest option will be returning each page number, and the whole file will be -1.

vkehfdl1 (changes requested):
Please fix untraceable parser output

bwook00:
add page and date column at `parser_node` result !

autorag/data/parse/base.py
```diff
@@ -32,10 +35,19 @@ def wrapper(
...
 
 	return wrapper
+
+
+def _add_date(result):
+	date_lst = [datetime.now()] * len(result[0])
+	result_lst = list(result)
+	result_lst.append(date_lst)
+	result_tuple = tuple(result_lst)
+	return result_tuple
```
vkehfdl1, line 53: I think you need to get a last modified datetime from each file.
There is `get_file_metadata` function at `autorag/data/utils/util.py`
Check that out

vkehfdl1 (changes requested):

bwook00:
change return "file_name" to "path"

vkehfdl1 (approved):
LGTM

PR #692, opened 2024-09-09
add override at Raw and Chunker from_parquet classmethod

### LINKED ISSUE #683: [Data Creation Refactoring] Merge multiple parsing results to one Raw. Load Raw instance to Chunker.
1. Raw needs some kind of merge with another Raw. Because sometimes there will be a multiple parsed result and wants to use at the same RAG pipeline. 
So make something like `Raw` plus override.
2. Use `Raw` instance at `Chunker` `__init__()` and load from file with new classmethod `from_parquet`

vkehfdl1:
close #683

### DIFF
--- a/autorag/chunker.py
+++ b/autorag/chunker.py
@@ -7,7 +7,6 @@ from typing import Optional
 
 import pandas as pd
 
-from autorag.data.beta.schema import Raw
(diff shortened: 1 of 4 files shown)


bwook00 (approved):
LGTM

PR #693, opened 2024-09-09
[Chunk] Add langchain chunk module

### LINKED ISSUE #682: Add langchain chunk module

bwook00:
close #682

Main review point !
1. add langchain embedding model at autorag/__inin__.py
     1-1. Langchain embedding model in init would be complicated and people(and "vkehfdl1") would hate it 😂. but as we are moving towards the future direction of the framework, we need to be able to import and use modules from Langchain, so I added it to init!!! 

2. delete "module_type" paramter at `get_start_end_idx`.

[Because]
2-1. The TextSplitter instance in Langchain doesn't have a class function that returns class_name or instance name
2-2. If `search_str` is not found in `original text` (= like Semantic Chunk), start_idx = -1 in `start_idx = original_text.find(search_str)`. So I add an if statement that can be used in this case.


autorag/data/__init__.py
```diff
@@ -25,6 +25,13 @@
 	SemanticDoubleMergingSplitterNodeParser,
 	SimpleFileNodeParser,
 )
+from langchain.text_splitter import (
+	RecursiveCharacterTextSplitter,
+	CharacterTextSplitter,
+	KonlpyTextSplitter,
+	SentenceTransformersTokenTextSplitter,
+)
+from langchain_experimental.text_splitter import SemanticChunker
```
vkehfdl1, line 34: I think we will not support semantic chunk at the first time, and make it on our own for index support?

autorag/data/chunk/langchain_chunk.py
```diff
@@ -0,0 +1,81 @@
...
+	results = loop.run_until_complete(process_batch(tasks, batch))
+
+	doc_id, contents, path, start_end_idx, metadata = (
+		list(chain.from_iterable(item)) for item in zip(*results)
+	)
+
+	return list(doc_id), list(contents), list(path), list(start_end_idx), list(metadata)
+
+
+async def langchain_chunk_pure(
```
vkehfdl1, line 51: There is no await.
No reason this is async.

vkehfdl1 (changes requested):

autorag/data/__init__.py
bwook00: Okay i agree

Issue #679, opened 2024-09-05
[Structure Refactoring] Refactor for efficient deploying


vkehfdl1:
**Is your feature request related to a problem? Please describe.**
The current deployment is so inefficient, because it loads each model and delete it at inference. 
It is good at optimization process, but not good at inference stage.

**Describe the solution you'd like**
![image](https://github.com/user-attachments/assets/e5954183-e4d8-4153-9112-3dcc6584e140)

## Refactor Modules - each including run.py
- [x] [Structure Refactoring] Retrieval
- [x] [Structure Refactoring] Query Expansion
- [x] [Structure Refactoring] Passage Reranker
- [x] [Structure Refactoring] Passage Augmenter
- [x] [Structure Refactoring] Passage Filter
- [x] [Structure Refactoring] Generator
- [x] [Structure Refactoring] Prompt Maker
- [x] [Structure Refactoring] Passage Compressor
- [ ] #728
- [ ] #729

## Extra
- [x] Refactoring Evaluator 
- [x] Refactoring Runner

## Test
- [x] Testing and resolve errors
- [x] Full run and real version test
- [x] Documentation

PR #697, opened 2024-09-10
Fix bug when use vllm in multi-gpu environment

### LINKED ISSUE #512: vLLM terminated unexpectedly at the multi-gpu environment
Hello,

While using the ELI5 and TriviaQA datasets from the Hugging Face library, I encountered errors related to missing documents that are not present in the corpus. I experienced a similar issue with the HotpotQA dataset but managed to resolve it by cleaning the mismatched documents.

However, when I switched to using the HotpotQA dataset, I observed some strange behavior with the VLLM module. Specifically, the process is terminated without any error messages.

Thank you in advance.

### LINKED ISSUE #691: [BUG] Does it work in a multi-GPU environment?
**Describe the bug**
When I try to infer llm locally, only the 0th GPU is used. Is there a guide to using multi GPU?

vkehfdl1:
close #691 
close #512

This is not complete fix for this error.
But we need to fix `run.py` and whole structure at #679 
So just fix this one from now.

### DIFF
--- a/autorag/nodes/generator/vllm.py
+++ b/autorag/nodes/generator/vllm.py
@@ -63,8 +63,10 @@ def make_vllm_instance(llm: str, input_args: Dict):
 
 	model_from_args = input_args.pop("model", None)
 	model = llm if model_from_args is None else model_from_args
-	init_params = inspect.signature(SamplingParams.__init__).parameters.values()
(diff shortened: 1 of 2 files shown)


bwook00 (approved):
LGTM

PR #713, opened 2024-09-16
fix vectordb score bug

vkehfdl1:
There was a crucial error at the vectordb.

When we use chromadb, the calculation method looks like this.

image

In this way, the lower score is better (more related to each other)
So, I resolve this with `1 - chroma_score`, getting original cosine similarity value.
I don't know why the score computation looks like this, but this is how it works.

### Effected modules

1. Ranking aware metrics (NDCG, mRR, mAP)
2. passage filters using scores

+ We have to measure again whole embedding benchmark


vkehfdl1:
@bwook00 Needs to be release to newer version. Since it is critical HotFix for the bug.

bwook00 (approved):
LGTM Let's go embedding benchmark again

PR #716, opened 2024-09-16
Ingest bm25_tokenizer and embedidng only in the strategy of other modules

### LINKED ISSUE #672: [Feature Request] Add guide for setup bm25_tokenizer in strategy of query_expansion
The user must set the value of bm25_tokenizer in startegy on the query_extend node to be the same as the bm25_tokenizer in bm25 on the retrieval node.    
If the bm25_tokenizer in retrieval node is missing any of the values in bm25_tokenizer in the query_extension node, an error is raised with: `bm25_path {bm25_path} does not exist. Please ingest first.`.    
Here is an example
```yaml
- node_line_name: pre_retrieve_node_line 
  nodes:
    - node_type: query_expansion
      strategy:
        metrics: [retrieval_f1, retrieval_recall, retrieval_precision]
        top_k: 5
        retrieval_modules:
          - module_type: bm25
            bm25_tokenizer: [ porter_stemmer, huggingface/new_model ]

- node_line_name: retrieve_node_line 
  nodes:
    - node_type: retrieval
      modules:
        - module_type: bm25
          bm25_tokenizer: [ porter_stemmer ]
```
There is no guidance or notice about this in the documentation, so I recommend adding a cautionary note about this situation.

There is no guide or notice about this in document. So it would be better to add guide the caution for this situation.

vkehfdl1:
close #672

- check test_evaluator.py for checking new test code. I edited sample.yaml for this feature.
- add un-commited api docs `.rst` files.
- Get `bm25_tokenizer` and `embedding_model` inside `retrieval_modules` and ingest it all


bwook00 (approved):
LGTM.
`find_key_values` func looks very useful

PR #750, opened 2024-09-26
add deepeval metrics

Eastsidegunn:
add metric and reform style


autorag/schema/metricinput.py
```diff
@@ -1,93 +1,117 @@
...
+					return False
+			except Exception:
+				return False
+
+		return True
+
+	def to_deepeval_testcase(self):
+		test_case = LLMTestCase(
+			input=self.query,
+			expected_output=self.generation_gt[0]
```
vkehfdl1, line 43: Do you not compute multiple generation_gt?

autorag/schema/metricinput.py
```diff
@@ -1,93 +1,117 @@
 from dataclasses import dataclass
 from typing import Optional, List, Dict, Callable, Any, Union
+from itertools import chain
 
 import numpy as np
 import pandas as pd
+from deepeval.test_case import LLMTestCase
```
vkehfdl1, line 7: import deepeval inside the `to_deepeval_testcase` 
Because I don't want deepeval import everytime use `MetricInput`.

vkehfdl1 (changes requested):
Did you use pre-commit and ruff format???
I think there is not formatted well or is it formatting all code again?

vkehfdl1:
@bwook00 

- Add deepeval faithfulness metric (see the prompt and the generation.py)
- Add `structure_output` functions at Generators. We can use structure output from our generator directly now. It uses more concise metric output that uses LLM judges. (We can update G-eval to support all generator modules)
- Do not run github action test on the all **legacy** qa data creation cases. Do not need it anymore.
- Many reformats. I don't know why this files reformatted. But maybe we have to make ruff format test on the CI/CD for checking it force (We can't see the difference non-ruff and yes-ruff easily)
- Delete RAGAS and guidance dependencies.

bwook00 (approved):
LGTM

vkehfdl1 (approved):

PR #766, opened 2024-09-29
Windows support (partially) AutoRAG

vkehfdl1:
- add `ignore_cleanup_errors` as True for working at Windows in every `tempfile.TemporaryDirectory()`
- add `delete=False` and close them after using it in every `tempfile.NamedTemporaryFile()`
- resolve `aiohttp` error at `JinaReranker`
- add documentation for windows install


bwook00 (approved):
LGTM

PR #784, opened 2024-10-03
[Parse Bug] Add non-table exists page to use clova.py

### LINKED ISSUE #779: [BUG] When there are no table in the page in clova, it occurs KeyError
**Describe the bug**

Go to `autorag/data/parse/clova.py` in line 103

```python
		table_html = json_to_html_table(resp_json["images"][0]["tables"][0]["cells"])
		page_text = extract_text_from_fields(resp_json["images"][0]["fields"])
```
When there are no key “tables”, it occurs key error.
It can be possible to put the pdf page that do not include table at all. Then it returns no table. 

**To Reproduce**
1. Use non-table exists page to use clova.py

**Expected behavior**
It must pass the table processing procedure.

bwook00:
close #779

### DIFF
--- a/autorag/data/parse/clova.py
+++ b/autorag/data/parse/clova.py
@@ -99,8 +99,10 @@ async def clova_ocr_pure(
 			raise RuntimeError(
 				f"Invalid response from Clova API: {resp_json['detail']}"
 			)
-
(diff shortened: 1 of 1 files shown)


vkehfdl1 (approved):
LGTM

PR #807, opened 2024-10-06
✨ feat: Added Docker push workflow, Dockerfile updates, and build script

hongsw:
add Docker Push workflow and stages for building and pushing Docker images with different variants for autoraghq/autorag, including handling production image tagging and updating Docker Hub description.


.github/workflows/docker-push.yml
```diff
@@ -0,0 +1,67 @@
+name: Docker Push
+
+on:
+  push:
+    branches: [ "main" ]
+    tags: [ "v*.*.*" ]
+  pull_request:
+    branches: [ "main" ]
```
vkehfdl1, line 8:
```yaml
      - name: Get changed files
        id: changed-files
        uses: tj-actions/changed-files@v44
      - name: Check for VERSION file change
        id: version_changed
        env:
          ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
        run: |
          echo "changed=false" >> $GITHUB_ENV
          if echo "${ALL_CHANGED_FILES}" | grep -q 'VERSION'; then
            echo "changed=true" >> $GITHUB_ENV
          fi
      - name: Build
        if: env.changed == 'true'
        run: |
          python -m build
      - name: Publish package distributions to PyPI
        if: env.changed == 'true'
        uses: pypa/gh-action-pypi-publish@release/v1
```


Here is the example of `publish.yml` to pypi.
We deploy our version when the `VERSION` is changed. 
I think the docker also build and push to the docker hub when the `VERSION` is changed.

vkehfdl1 (changes requested):
Plus I will add secrets when you send that privately

.github/workflows/docker-push.yml
```diff
@@ -0,0 +1,92 @@
...
```
vkehfdl1, line 8: I think you have to delete this after doing experiment on the github actions
Just trigger when push to branch main
and do not push to docker hub when no version changed

vkehfdl1 (changes requested):

.github/workflows/docker-push.yml
hongsw: agree

vkehfdl1 (approved):
LGTM

PR #809, opened 2024-10-06
Add VoyageAI Reranker module

### LINKED ISSUE #793: [Feature Request] New Voyageai-Reranker

bwook00:
close #793 

I wanted to make `mock_voyageai_reranker` a mock function that returns a `RerankingObject`(Their own class type) in my test code, but there are too many things to set up to create a `RerankingObject` instance, so I just made the mock def simple.

The test passed perfectly


autorag/nodes/passagereranker/voyageai.py
```diff
@@ -0,0 +1,106 @@
...
+	:param voyage_client: The Voyage Client to use for reranking
+	:param model: The model name for VoyageAI rerank
+	:param query: The query to use for reranking
+	:param documents: The list of contents to rerank
+	:param ids: The list of ids corresponding to the documents
+	:param top_k: The number of passages to be retrieved
+	:param truncation: Whether to truncate the input to satisfy the 'context length limit' on the query and the documents.
+	:return: Tuple of lists containing the reranked contents, ids, and scores
+	"""
+	rerank_results = voyage_client.rerank(
```
vkehfdl1, line 91: Is there any async operation at the voyage_client?

I bet they are not local model since it uses their own API key. 
Please find the async operation, not just for-loop request. 
It is I/O blocking operation and effects so much time to overall AutoRAG services.

vkehfdl1 (changes requested):

PR #812, opened 2024-10-07
Properly truncate to 8000 tokens when we use OpenAI Embeddings

### LINKED ISSUE #811: [BUG] OpenAIEmbedding did not truncate as openai_truncate_by_token
**Describe the bug**
isinstance(embedding_model, OpenAIEmbedding) => this will be False

You can use this

isinstance(embedding_model._instance, OpenAIEmbedding)

**Code that bug is happened**
If applicable, add the code that bug is happened.
(Especially, your AutoRAG YAML file or python codes that you wrote)

 - OS: [e.g. Windows, Linux, MacOS]
 - Python version [e.g. 3.10]

**Additional context**

vkehfdl1:
close #811

It will be great we can set truncate limit as parameter => It can be a new issue.

- Call embedding model before use
- Adjust embedding token limit to 8000

### DIFF
--- a/autorag/data/legacy/qacreation/base.py
+++ b/autorag/data/legacy/qacreation/base.py
@@ -166,7 +166,7 @@ def make_qa_with_existing_qa(
 		content_size = len(corpus_df)
 
 	logger.info("Loading local embedding model...")
-	embeddings = autorag.embedding_models[embedding_model]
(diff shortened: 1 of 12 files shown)


bwook00 (approved):
LGTM

PR #835, opened 2024-10-12
change stream to astream, and add non-async stream function

### LINKED ISSUE #834: Change stream => astream and make stream function which is not asynchronous generator
For more implementation of features

vkehfdl1:
close #834

This is slight modification for the our M0 project.


bwook00 (approved):
LGTM

PR #837, opened 2024-10-12
Change recency filter parameter name to threshold_datetime from threshold

### LINKED ISSUE #822: [Bug] Restrict 'threshold' config to numbers only for module_type
> Restrict 'threshold' config to numbers only for module_type 

 _Originally posted by @hongsw in [#821](https://github.com/Marker-Inc-Korea/AutoRAG/issues/821#issuecomment-2403716359)_

## Is your feature request related to a problem? Please describe.
Currently, the `threshold` configuration for `module_type` accepts any type of value, which can lead to validation errors and potential runtime issues. Specifically, we've observed instances where date strings (e.g., "2015-01-01T00:00:00.000Z") are being used instead of numerical values. This inconsistency can cause confusion for users and may result in unexpected behavior in our application.

For example, we've seen the following error:

```
TYPE must be number

  216 |             {
  217 |               "module_type": "recency_filter",
> 218 |               "threshold": "2015-01-01T00:00:00.000Z"
      |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^ 👈🏽  type must be number
```

This indicates that a string is being used where a number is expected.

## Describe the solution you'd like
We propose to modify the JSON schema to explicitly restrict the `threshold` configuration for `module_type` to accept only numerical values. This change should:

1. Update the JSON schema to specify that `threshold` must be a number.
2. Provide clear error messages when non-numerical values are used for `threshold`.
3. Update any relevant documentation to clearly state that `threshold` should be a number.

## Describe alternatives you've considered
1. Allow both numbers and date strings, and convert date strings to numerical timestamps internally:
   - Pros: More flexible for users who prefer working with dates
   - Cons: Increases complexity and may lead to inconsistent usage

2. Use a separate configuration for date-based thresholds:
   - Pros: Clear separation between numerical and date-based thresholds
   - Cons: Requires more configuration options, potentially complicating the schema

## Additional context
This change will improve the consistency of our configuration schema and reduce errors caused by incorrect data types. It's part of our ongoing effort to make the configuration process more robust and user-friendly.

Example of the desired schema change:

```json
{
  "properties": {
    "module_type": {
      "type": "object",
      "properties": {
        "threshold": {
          "type": "number"
        }
      }
    }
  }
}
```

This change will ensure that only numerical values are accepted for the `threshold` configuration.

vkehfdl1:
close #822

### DIFF
--- a/sample_config/rag/full.yaml
+++ b/sample_config/rag/full.yaml
@@ -88,7 +88,7 @@ node_lines:
         - module_type: similarity_percentile_cutoff
           percentile: 0.6
         - module_type: recency_filter
-          threshold: 2015-01-01
+          threshold_datetime: 2015-01-01 3:45:07
         - module_type: threshold_cutoff
           threshold: 0.85
         - module_type: percentile_cutoff
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -10,8 +10,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
     - name: Upgrade pip
-      run: | 
+      run: |
         python -m pip install --upgrade pip
     - name: Install gcc
       run: |
--- a/autorag/deploy/api.py
+++ b/autorag/deploy/api.py
@@ -145,7 +145,7 @@ class ApiRunner(BaseRunner):
 						# Start streaming of the result
 						assert len(previous_result) == 1
 						prompt: str = previous_result["prompts"].tolist()[0]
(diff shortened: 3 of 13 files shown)


hongsw:
Full.yaml file must update. @vkehfdl1

sample_config/rag/full.yaml
```diff
@@ -88,7 +88,7 @@ node_lines:
         - module_type: similarity_percentile_cutoff
           percentile: 0.6
         - module_type: recency_filter
-          threshold: 2015-01-01
+          threshold_datetime: 2015-01-01 3:45:07
```
hongsw, line 91: 😊 thanks

hongsw (approved):

PR #850, opened 2024-10-16
Fix README.md

Jake-Song:
In my understanding, raw data is parsed data right? it seems like incorrect to me.


vkehfdl1 (approved):
Thanks:)
Silly mistake from me haha

PR #876, opened 2024-10-22
Story/cip 64/accept chat summary

Raihan-Seraj:
Added capability to incorporate chat summary

### DIFF
--- a/autorag/nodes/generator/exo_openailike_llm.py
+++ b/autorag/nodes/generator/exo_openailike_llm.py
@@ -1,0 +1,215 @@
+import logging
+from typing import List, Tuple
+
+import pandas as pd
+import tiktoken
+from openai import AsyncOpenAI
+from tiktoken import Encoding
+
+from autorag.nodes.generator.base import BaseGenerator
+from autorag.utils.util import (
+	get_event_loop,
+	process_batch,
+	pop_params,
+	result_to_dataframe,
+)
+
+logger = logging.getLogger("AutoRAG")
+
+MAX_TOKEN_DICT = {  # model name : token limit
+	"llama-3.1-8b": 8_192,
+}
+
+
+class ExoOpenAILikeLLM(BaseGenerator):
+	def __init__(self, project_dir, llm: str, batch: int = 16, *args, **kwargs):
+		super().__init__(project_dir, llm, *args, **kwargs)
+		assert batch > 0, "batch size must be greater than 0."
+		self.batch = batch
+
+		client_init_params = pop_params(AsyncOpenAI.__init__, kwargs)
+		self.client = AsyncOpenAI(**client_init_params)
+
+		if self.llm.startswith("o1"):
+			self.tokenizer = tiktoken.get_encoding("o200k_base")
+		else:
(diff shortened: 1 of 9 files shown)


test.yaml
```diff
@@ -0,0 +1,103 @@
...
+             model: gpt-4
+           - metric_name: bert_score
+             lang: en
+             batch: 32
+        speed_threshold: 10
+        token_threshold: 4000
+      modules:
+        - module_type: openai_llm
+          llm: [gpt-4, gpt-4o]
+          temperature: [0.1, 1.0]
```
vkehfdl1, line 103: add new line please

autorag/nodes/promptmaker/fstring.py
```diff
@@ -26,24 +26,25 @@ def _pure(
...
 		:return: Prompts that are made by f-string.
 		"""
 
 		def fstring_row(
-			_prompt: str, _query: str, _retrieved_contents: List[str]
+			_prompt: str, _query: str, _retrieved_contents: List[str], _chat_summary: str
 		) -> str:
 			contents_str = "\n\n".join(_retrieved_contents)
-			return _prompt.format(query=_query, retrieved_contents=contents_str)
+			return _prompt.format(query=_query, retrieved_contents=contents_str, summary=_chat_summary)
```
vkehfdl1, line 43: Also this need to be an optional.

Why don't we get prompt_with_chat as parameter?
If there is no chat_summary input, just use ordinary prompt.
But, when chat_summary is in the input, use prompt_with_chat.

It could be a easy solution without making other chat summary modules in the system.

test.yaml
```diff
@@ -0,0 +1,103 @@
+node_lines:
```
vkehfdl1, line 1: Plus can you move this YAML file to `sample_config` folder?

autorag/nodes/generator/exo_openailike_llm.py
```diff
@@ -0,0 +1,215 @@
+import logging
```
vkehfdl1, line 1: Can you tell me why you made `exo_openailike_llm`??
It is better just use LlamaIndexLLM `openailike`??
Plus, since we add chat history at prompt maker, it has no effect to the generator. And we can implement chat history without tackling with generator modules.

vkehfdl1 (changes requested):
Thank you! Great idea to implement chat history. Never thought about just adding it to prompt maker. Think about it can be a total new challenge to optimize chat history data.

+ One small note: Please run a linter. You can use ruff linter use `pre-commit install` at our repository.

autorag/nodes/generator/exo_openailike_llm.py
dcampbelbraxted: Hey folks, some of this wasn't meant to be pushed yet - we're still testing! We'll roll back the PR, but implement the above comments. 

FYI LLamaIndexLLM openailike seems to not play well with Exo LLM endpoints due to something about the messages object passing. Re-writing the OpenAI lib to interact solved this issue, to a certain degree. 

Cheers,
Daniel & Team

PR #920, opened 2024-11-02
Fix: added table_html variable initialization

effortprogrammer:
While performing table hybrid parse with naver ocr, there was error: 

```python
 UnboundLocalError: local variable 'table_html' referenced before assignment         
```

To fix this error, I added table_html variable initialization to solve this error. After that, code works as expected.


bwook00 (approved):
@effortprogrammer 
Thanks for your contribution!
That is our huge mistake 😂
LGTM

PR #931, opened 2024-11-10
just return original texts when there is no corresponding tokenizer a…

### LINKED ISSUE #924: [BUG] Upstage Embedding not working
At vectordb.py, llamaindex UpstageEmbedding class is instance of OpenAIEmbedding.
So it tires to truncate token to 8000, but tiktoken cannot tokenize it because solar embedding model is not capable of tokenize in tiktoken. 

Add model_name condition at truncate openai token also.

+ 
Just change colab demo to BAAI

vkehfdl1:
…t openai_truncate_by_token

close #924


bwook00 (approved):
LGTM

PR #945, opened 2024-11-13
Fix extracting vectordb from config.yaml

### LINKED ISSUE #942: [BUG] The vectordb setting is always applied as default when validating
**Describe the bug**
Event I configure vectordb, always set as default when validating

**To Reproduce**
1. Configure yaml with custom vectordb configuration
2. Run validation
3. See error

**Expected behavior**
Ingest vectordb with my configuration

**Full Error log**
not error

**Code that bug is happened**
`config_dict` is full dict of config.yaml, so it only has **node_lines** as key. 
vectordb_list always becomes empty list because `config_dict` has to "vectordb" key.  
If that's not what you intended, it need to be fixed it.
```python
def load_all_vectordb_from_yaml(
    yaml_path: str, project_dir: str
) -> List[BaseVectorStore]:
    config_dict = load_yaml_config(yaml_path)
    vectordb_list = config_dict.get("vectordb", [])
    if len(vectordb_list) == 0:
        chroma_path = os.path.join(project_dir, "resources", "chroma")
        return [
            load_vectordb(
                "chroma",
                client_type="persistent",
                embedding_model="openai",
                collection_name="openai",
                path=chroma_path,
            )
        ]
...
```
**Suggestion**
We need to ingest vectordb configured in retrieval node

rjwharry:
close #942

### DIFF
--- a/autorag/evaluator.py
+++ b/autorag/evaluator.py
@@ -21,6 +21,7 @@ from autorag.nodes.retrieval.vectordb import (
 )
 from autorag.schema import Node
 from autorag.schema.node import (
+	extract_vectordb_from_nodes,
 	module_type_exists,
 	extract_values_from_nodes,
 	extract_values_from_nodes_strategy,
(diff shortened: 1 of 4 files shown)


rjwharry:
I didn't realize that the way of configuring vectordb is changed. Let me close this issue

PR #949, opened 2024-11-15
Add Weaviate VectorDB

### LINKED ISSUE #943: [Feature Request] Add Weaviate vector db

bwook00:
close #943 

Review Point
- I haven't found a way to utilize query_embeddings in Query all at once. so I implemented it to use for-loop
(I searched Official Docs, Blog, Langchain, and Llama Index for a way to do this, but couldn't find one 🥲)

### DIFF
--- a/autorag/vectordb/weaviate.py
+++ b/autorag/vectordb/weaviate.py
@@ -46,7 +46,7 @@ class Weaviate(BaseVectorStore):
 		else:
 			raise ValueError(
 				f"client_type {client_type} is not supported\n"
-				"supported client types are: local_docker, cloud"
+				"supported client types are: docker, cloud"
 			)
 		if similarity_metric == "cosine":
 			distance_metric = wvc.config.VectorDistances.COSINE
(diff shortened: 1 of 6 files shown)


autorag/vectordb/weaviate.py
```diff
@@ -0,0 +1,167 @@
...
+			)
+		elif client_type == "cloud":
+			self.client = weaviate.connect_to_weaviate_cloud(
+				cluster_url=url,
+				auth_credentials=Auth.api_key(api_key),
+			)
+		else:
+			raise ValueError(
+				f"client_type {client_type} is not supported\n"
+				"supported client types are: local_docker, cloud"
```
vkehfdl1, line 49: I think
"supported client types are docker, cloud"

Not "local_docker"

vkehfdl1 (changes requested):
I think you did good job at importing vectors. 
They are using gRPC to batch importing. 

One more thing:
Did you check the distance metric scores?? 
https://weaviate.io/developers/weaviate/config-refs/distances

I think cosine it is better to use 1 - distance.
Others looks okay

bwook00:
image

cosine distance is 0: identical vectors, 2: Opposing vectors, so I wanted to make the Scale of the Score be 0-2.
Is this too confusing? @vkehfdl1

vkehfdl1 (approved):
LGTM

PR #975, opened 2024-11-21
Add env variable endpoints for managing env variable

### LINKED ISSUE #974: [API] Add endpoint for env variable management
GET /env => return all env variable key

DELETE /env/<string:key> => delete the certain key as a env variable.

We will manage with '.env' file comes with app.py file. 
Load it when the api-key needed endpoint starting. (parse, chunk, qa creation, validation, evaluate, chat)

vkehfdl1:
close #974

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -1197,6 +1197,15 @@ async def get_environment_variable(key: str):
         return {"error": f"Internal server error: {str(e)}"}, 500
 
 
+@app.route("/env", methods=["GET"])
+async def get_all_env_keys():
+    try:
+        envs = dotenv_values(ENV_FILEPATH)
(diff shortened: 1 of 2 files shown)


api/app.py
```diff
@@ -1187,6 +1200,30 @@ async def get_environment_variable(key: str):
         return {"error": f"Internal server error: {str(e)}"}, 500
 
 
+@app.route("/env", methods=["GET"])
+async def get_all_env_keys():
+    try:
+        envs = dotenv_values(ENV_FILEPATH)
```
hongsw, line 1206: only getting ENV value which we saved it only. good point

hongsw (dismissed):

PR #988, opened 2024-11-23
Replace to local embeddings at the gpu sample config YAML files

### LINKED ISSUE #986: [Sample Config Fix] Update non_gpu YAML files to use local embeddings.
After v0.3.8 we have to configure embeddings at seperated vecotrdb settings.

vkehfdl1:
close #986


bwook00 (approved):
LGTM

PR #1044, opened 2024-12-09
change file type logic

### LINKED ISSUE #1040: [HotFix] I think the logic at the file type reduction is wrong at the parser
**Describe the bug**
parse/run.py 56:61

```python
# create a list of only those file_types that are in file_types but not in set_file_types
missing_file_types = list(file_types - set_file_types)
if list(set_file_types - file_types):
	raise ValueError(
		f"File types {list(set_file_types - file_types)} are not in the data path.”
	)
```
I don’t know why the hell we raise an error when `set_file_types - file_types`. I bet people wants just ignore the error when this kind of error occured. 

**To Reproduce**
When you put file_type to pdf, csv, html, md…
Only put pdf files in the folder.
You can get an `ValueError: File types ['csv', 'json', 'xml', 'html', 'md'] are not in the data path.`

**Expected behavior**
I think it supposed to act like this.

### 1. If there are no file type, even user write an file_type in the YAML file.

Just ignore it. Don’t load the module that doesn’t have any file is quite good idea.

 ### 2. If there is a new file type, even the user doesn’t write an file_type in the YAML file.

Raise an warning, not error. I bet skip it is okay. But just raise an warning for user will notice it.

bwook00:
close #1040 

+ upgrade rc3


autorag/data/parse/run.py
```diff
@@ -56,9 +56,10 @@ def run_parser(
 		# create a list of only those file_types that are in file_types but not in set_file_types
 		missing_file_types = list(file_types - set_file_types)
 		if list(set_file_types - file_types):
-			raise ValueError(
-				f"File types {list(set_file_types - file_types)} are not in the data path."
-			)
+			for module, module_param in zip(modules, module_params):
+				if module_param["file_type"] in list(set_file_types - file_types):
+					modules.remove(module)
+					module_params.remove(module_param)
```
vkehfdl1, line 62: When you do this, I think there will be the problem at the for-loop

vkehfdl1 (changes requested):

vkehfdl1 (approved):
LGTM

PR #1071, opened 2025-01-19
add llama index's ollama embedding and fix type hint of embedding_model

### LINKED ISSUE #1070: [Feature Request] Add Ollama Embedding
Not supporting ollama for embedding model.
It may related to #985 

Fix the bug and add ollama for embedding model

**Additional context**

rjwharry:
Resolve #1070 
- Add llama index's OllamaEmbedding class.
- Separate original `add_from_dict` function to `add_from_dict` and `add_from_list`. Because of `get_param_combinations`, modules will get embedding_model config as dictionary not list. But vectordb will get it as list. So I separate to two functions to handle both cases.
- Unify the way of loading embedding model.

### DIFF
--- a/autorag/data/chunk/base.py
+++ b/autorag/data/chunk/base.py
@@ -4,7 +4,7 @@ from typing import Tuple, List, Dict, Any
 
 import pandas as pd
 
-from autorag.embedding.base import embedding_models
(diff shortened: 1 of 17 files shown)


vkehfdl1 (approved):
LGTM Thanks:)

PR #1103, opened 2025-03-28
Correct Japanese prompts to be more natural.

sappho192:
Hi,

I've made some corrections to Japanese prompts to be more natural and robust to make the QA dataset.

It seems that Japanese prompt is machine-translated from Korean or English prompt, and the prompt had some typos and misleading expressions.

As a result, current Japanese prompt generates query that asking about the file name, which should be prohibited according to other languages' prompt.
(ファイル名は何ですか？ = What is the file name?)
image

By the way, thank you for releasing this awesome project. AutoRAG is straightforward and easy to use 🤗


vkehfdl1:
@sappho192 
Thank you for your contribtuion!

PR #1160, opened 2025-11-05
Add the vllm embedding model

### LINKED ISSUE #1157: Support vLLM embedding
vLLM supports an amazing embedding method

vkehfdl1:
close #1157

### DIFF
--- a/autorag/autorag/embedding/vllm.py
+++ b/autorag/autorag/embedding/vllm.py
@@ -46,22 +46,10 @@ class VllmEmbedding(MultiModalEmbedding):
 		description="Holds any model parameters valid for `vllm.LLM` call not explicitly specified.",
 	)
 
-	retry_cnt: int = Field(
-		default=3,
-		description="Number of retries for embedding in case of failure.",
-	)
-	wait_multiplier: int = Field(
-		default=1,
-		description="The multiplier factor to use for exponential backoff during retries.",
-	)
-	wait_min: int = Field(
-		default=4,
-		description="Minimum wait time (in seconds) between retries.",
-	)
-	wait_max: int = Field(
-		default=10,
-		description="Maximum wait time (in seconds) between retries.",
-	)
+	retry_cnt: int = 3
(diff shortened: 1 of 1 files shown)


autorag/autorag/embedding/vllm.py
```diff
@@ -0,0 +1,251 @@
...
+		Generates Embeddings for Query Asynchronously.
+
+		Args:
+		    query (str): Query text/sentence
+
+		Returns:
+		    List[float]: numpy array of embeddings
+
+		"""
+		return self._get_query_embedding(query)
```
bwook00: I think its not support async because the `_get_query_embedding` function is not async function. 
Shouldn't we use awiat as follows? or Vllm doesn't support any aysnc embed function?
```
loop = asyncio.get_event_loop()
return await loop.run_in_executor(None, self._get_query_embedding, query)
```

autorag/autorag/embedding/vllm.py
```diff
@@ -0,0 +1,251 @@
...
+		import gc
+
+		if torch.cuda.is_available():
+			gc.collect()
+			torch.cuda.empty_cache()
+			torch.cuda.synchronize()
+
+	@retry(
+		stop=stop_after_attempt(3),
+		wait=wait_exponential(multiplier=1, min=4, max=10),
```
bwook00: I think using a constant is better than a hardcoded int.

bwook00 (changes requested):
I think we have to use `AsyncLLMEngine` from vllm to support async func

https://docs.vllm.ai/en/v0.6.2/dev/engine/async_llm_engine.html

vkehfdl1:
> I think we have to use `AsyncLLMEngine` from vllm to support async func
> 
> https://docs.vllm.ai/en/v0.6.2/dev/engine/async_llm_engine.html

https://docs.vllm.ai/en/latest/api/vllm/v1/engine/async_llm.html#vllm.v1.engine.async_llm.AsyncLLM

Latest doc is here.

PR #1180, opened 2025-11-11
Add PostgreSQL Docker folder and its test code at the conftest.py and GitHub Actions

### LINKED ISSUE #1165: PostgreSQL 세팅 및 도커 설정 (초기 스키마 설정 등)

bwook00:
close #1165

The test was successful(?)

image


tests/conftest.py
```diff
@@ -1,5 +1,34 @@
...
+import time
+import pytest
 
 
 def pytest_sessionstart(session):
-    os.environ["BM25"] = "bm25"
+	os.environ["BM25"] = "bm25"
+
+
+@pytest.fixture(scope="session", autouse=True)
```
vkehfdl1: The PR uses a pytest fixture in the `_postgres_healthcheck` but does not return any instance in the function. This is bad because the `pytest` fixture should return the instance for reusability during testing. I recommend deleting the `pytest` fixture decorator because this function should exist only for the health check of PostgreSQL.

bwook00:
@vkehfdl1 The official Docker PostgreSQL image will run all *.sh and *.sql scripts in its /docker-entrypoint-initdb.d directory automatically when it starts.

PR #1201, opened 2026-02-28
feat: add Azure OpenAI support with docs and sample config

### LINKED ISSUE #749: [Feature Request] Add azure api docs and tutorial (+ sample config files)
I see many people using Azure OpenAI, because whole company using Azure only. 

Test Azure openai system in AutoRAG, and find out what will be needed.

I think there are some modules that can only use OpenAI (like g_eval..?)
It can be an issue;;

**Additional context**

ayzmkk-rongbi:
- Add azure_openai as a supported LLM provider in generator_models
- Create sample config YAML for Azure OpenAI (simple_azure_openai.yaml)
- Add comprehensive Azure OpenAI integration documentation
- Update local_model.md to include Azure OpenAI in the LLM table
- Add azure_openai.md to docs index

End to End test:

image

Closes #749


JiwaniZakir (commented):
The `AzureOpenAI` class is registered directly in `generator_models["azure_openai"] = AzureOpenAI` (in `autorag/__init__.py`), while the docs' embedding example wraps the class in `LazyInit`. If other parts of the framework expect all generator model values to be `LazyInit` instances rather than raw classes, this inconsistency could cause runtime failures when the azure_openai generator is selected. It's worth verifying that the instantiation path in `llama_index_llm` module handles a raw class the same way it handles a `LazyInit`-wrapped one.

There's also a parameter name inconsistency in the docs: the query expansion example under "Using Azure OpenAI with Query Expansion" uses `max_token: 64` (singular), while the parameters table lists `max_tokens` (plural). Depending on how kwargs are forwarded to the LlamaIndex class, the singular form would silently be ignored, leading to unexpectedly long responses.

Finally, the "Option 1" embedding approach sets `OPENAI_API_KEY` and `OPENAI_API_BASE` to Azure values globally, which would break any standard OpenAI usage running in the same process — this side-effect deserves an explicit warning in the docs rather than the current "Recommended" label.

ayzmkk-rongbi:
Nice work on this! Reviewed the changes — the `__init__.py` registration is solid, raw class is consistent with how `openai`, `bedrock`, etc. are all done.

Pushed a small fix commit (b14e2e1) for two doc issues I spotted:

1. **`max_token` → `max_tokens`** in the query expansion example — the singular form gets silently dropped by `pop_params()`, so it wouldn't actually limit the response length.
2. **Embedding Option 1** was labeled "Recommended" but setting `OPENAI_API_KEY`/`OPENAI_API_BASE` globally would break any standard OpenAI calls in the same process. Added a warning and moved the "Recommended" label to Option 2 (the `LazyInit` approach) which is safer.

vkehfdl1:
Superseded by reviewed replacement #1340, which preserves this PR’s implementation, ports it to the current legacy/ layout, corrects the unsupported Azure embedding claims in the docs/sample, and passed manual provider instantiation plus refreshed CI.

Issue #1162, opened 2025-11-05
Add data creation pipeline for making visual image based QA


vkehfdl1:

Rishabh-git10:
Hi, I'd like to pick up this issue.
Since the current pipeline relies on pd dfs and is strictly optimized for text extraction; I was thinking of the following approach to support VLMs for visual QA creation :

1. Updating the `parser_node` wrapper and schemas to support an `image_path` (or `image_base64`) column alongside the standard `texts` column.
2. Adding vision-compatible generators (`openai_vlm_gen_query` or similar) that can pass the stored image data to the VLM API to populate the `query` column.
3. Integrating VLM calls to generate the `generation_gt` answers based on the visual context.

Does this align with the project's roadmap? Also, do you have a preference for storing local file paths vs base64-encoded strings directly in the DFs?

PR #1202, opened 2026-02-28
[WIP] feat: Visual QA data creation PoC (#1162)

Rishabh-git10:
Relates to #1162 

Hey @vkehfdl1, since the issue didn't have a description, I threw together a quick draft PR as a PoC to make sure we're aligned on the architecture for visual QA creation before I build out the whole pipeline.

Here's the approach I'm taking:
* **Data layer (`schema.py`):** Updated `QA.to_parquet` to persist an `image_path` column. Assuming it's better to pass local paths through the pipeline rather than storing raw base64 in the DataFrames to avoid memory bloat.
* **Generator (`openai_vlm_gen_query.py`):** Stubbed out a VLM query generator. It dynamically converts the local `image_path` to base64 at the API boundary right before sending it to OpenAI.

_Quick note on the model: I defaulted to `gpt-4o-2024-08-06` to match the existing text modules and keep tests passing. Given OpenAI's transition to `gpt-5.2` this month, let me know if you'd rather bump the default specifically for these vision nodes._

Still left on the to-do list before this is ready for review:
* Update `parse/base.py` for multimodal file ingestion
* Add the VLM `generation_gt` module
* Add tests and mock the new API calls

Let me know if this data flow and file structure work for you, and I'll finish up the rest of the implementation.

### DIFF
--- a/autorag/data/qa/query/openai_vlm_gen_query.py
+++ b/autorag/data/qa/query/openai_vlm_gen_query.py
@@ -1,0 +1,75 @@
+import base64
+import itertools
+from typing import Dict, List
+
+from llama_index.core.base.llms.types import ChatMessage, MessageRole
+from openai import AsyncClient
+from pydantic import BaseModel
+
+from autorag.data.qa.query.prompt import QUERY_GEN_PROMPT
+
+
+class Response(BaseModel):
+	query: str
+
+
+def encode_image(image_path: str) -> str:
+	with open(image_path, "rb") as image_file:
+		return base64.b64encode(image_file.read()).decode("utf-8")
+
+
+# Single hop Visual QA generation OpenAI
+async def query_gen_vlm_openai_base(
+	row: Dict,
+	client: AsyncClient,
+	messages: List[ChatMessage],
+	model_name: str = "gpt-4o-2024-08-06",
+):
+	context = list(itertools.chain.from_iterable(row.get("retrieval_gt_contents", [])))
+	context_str = "Text:\n" + "\n".join(
+		[f"{i + 1}. {c}" for i, c in enumerate(context)]
+	)
+	user_prompt = f"{context_str}\n\nGenerated Question from the Text and Image:\n"
+
+	openai_messages = []
+	for msg in messages:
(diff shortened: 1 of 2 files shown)


JiwaniZakir (commented):
In `openai_vlm_gen_query.py`, the MIME type is hardcoded as `image/jpeg` in the base64 URL (`data:image/jpeg;base64,...`), but `image_path` could point to a PNG, WebP, GIF, or other format — this will cause silent failures or incorrect rendering with non-JPEG images. The MIME type should be derived from the file extension using something like `mimetypes.guess_type(row["image_path"])`.

Additionally, `encode_image` performs a blocking file read inside the `async` function `query_gen_vlm_openai_base`, which will block the event loop when processing large images or many concurrent rows. This should use `asyncio` file I/O (e.g., `aiofiles`) instead.

The `vlm_factoid_query_gen` function accesses `QUERY_GEN_PROMPT["factoid_single_hop"][lang]` without guarding against unsupported `lang` values — a `KeyError` here will surface as an unhelpful traceback rather than a meaningful error message about which language is unsupported.

The change in `schema.py` to conditionally include `image_path` in `columns_to_save` is clean and backward-compatible, but there's no corresponding handling in whatever deserialization or loading path reads back the saved parquet — if consumers of this file don't expect `image_path` to be optionally present, that could cause issues downstream.

Rishabh-git10:
Thanks for the review, @JiwaniZakir.

1. **MIME type & blocking I/O:** Fixed. Just pushed a patch adding `aiofiles` and dynamic MIME resolution.
2. **`lang` guard & `schema.py`:** Left these out intentionally to mirror the repo's existing text pipeline patterns to keep the PoC backward-compatible.

vkehfdl1:
Closing because this is explicitly an incomplete PoC and is not usable through the legacy QA creation surface: multimodal ingestion, generation_gt support, module registration/export, deserialization handling, language validation, and tests are all missing. It also targets the old pre-legacy/ layout and an older OpenAI beta API path. A complete visual-QA design should be proposed against the current architecture rather than merging this partial pipeline.

PR #1236, opened 2026-04-17
fix: hotfixes for v0.3.22 regressions (openai, nvidia reranker, chunk embedding, llama-index logprobs)

### LINKED ISSUE #1235: [BUG] data/qa openai helpers still call the removed client.beta.chat.completions.parse API
**Describe the bug**
PR #1131 migrated `autorag/nodes/generator/openai_llm.py` off the now-removed `client.beta.chat.completions.parse` API, but the QA data-creation pipeline still uses it in seven places across five files:

- `autorag/data/qa/query/openai_gen_query.py:30, 89`
- `autorag/data/qa/filter/dontknow.py:80`
- `autorag/data/qa/filter/passage_dependency.py:40`
- `autorag/data/qa/generation_gt/openai_gen_gt.py:28`
- `autorag/data/qa/evolve/openai_query_evolve.py:33, 75`

With `openai>=2.8.0` (the range pinned in `pyproject.toml`), the `AsyncOpenAI` client no longer exposes the `.beta.chat.completions.parse` entry point, so every call above raises `AttributeError` and the entire QA-generation workflow (`factoid_query_gen`, `make_basic_gen_gt`, `dontknow_filter_openai`, `passage_dependency_filter_openai`, `query_evolve_openai_base`, etc.) fails on first use.

**To Reproduce**

1. `pip install AutoRAG==0.3.22` (or install from current `main`).
2. Run any QA creation helper that depends on OpenAI structured output:
   ```python
   import asyncio
   from openai import AsyncClient
   from autorag.data.qa.query.openai_gen_query import factoid_query_gen

   async def main():
       client = AsyncClient()
       row = {"retrieval_gt_contents": [["Some text about AutoRAG."]]}
       await factoid_query_gen(row, client)

   asyncio.run(main())
   ```
3. Observe an `AttributeError` pointing at `client.beta.chat.completions.parse`.

**Expected behavior**
The QA pipeline should call the same `client.responses.parse(input=..., text_format=...)` API that the generator module already migrated to, and read the structured object from `completion.output_parsed`.

**Full Error log**
```
AttributeError: 'AsyncOpenAI' object has no attribute 'beta'
```
(Exact wording depends on the openai SDK version; the common cause is the same: `beta.chat.completions.parse` has been removed.)

**Code that bug is happened**
The six call sites listed above; they all share this pattern:

```python
completion = await client.beta.chat.completions.parse(
    model=model_name,
    messages=to_openai_message_dicts(messages),
    response_format=Response,
)
row["query"] = completion.choices[0].message.parsed.query
```

`autorag/nodes/generator/openai_llm.py::get_structured_result` already shows the post-migration pattern:
```python
response = await self.client.responses.parse(
    model=self.llm,
    input=parse_prompt(prompt),
    text_format=output_cls,
    **kwargs,
)
return response.output_parsed
```

 - OS: macOS 15.3 (reproduces on Linux as well)
 - Python version: 3.12

**Additional context**
This is follow-up cleanup of PR #1131; that PR only covered the generator module. I have a focused fix ready that applies the identical migration to the seven call sites above (no behavioural change beyond the API surface) — see the linked PR.

shaun0927:
## Summary
This PR bundles small, independent fixes for regressions that surface on a plain install of the current `main` / v0.3.22. Each commit is isolated and cherry-pickable so maintainers can take whatever subset makes sense.

### 1. `OpenAILLM.__init__` crashes for any model outside `MAX_TOKEN_DICT` (Critical)
`autorag/nodes/generator/openai_llm.py`

`MAX_TOKEN_DICT.get(self.llm) - 7` is evaluated before the `None` guard, so any model alias missing from the dictionary (e.g. `"gpt-4o-latest"`, a freshly-released alias, or a typo) crashes the constructor with:

```
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```

The downstream `if self.max_token_size is None` raise was dead code for that path.

**Fix**: look up the limit first, raise `ValueError` with the list of supported models if it is missing, subtract `7` only on success. While touching the file, drop `gpt-5` from the `get_result_reasoning` error message (the guard above it only checks `o1`/`o3`/`o4` — `gpt-5` has its own `get_result_gpt_5` branch, so the message was misleading).

### 2. `data/qa/*` still calls the removed `client.beta.chat.completions.parse` (Critical)
PR #1131 migrated `autorag/nodes/generator/openai_llm.py` off the removed beta namespace, but the QA data-creation pipeline kept seven call sites across five files:

- `autorag/data/qa/query/openai_gen_query.py:30, 89`
- `autorag/data/qa/evolve/openai_query_evolve.py:33, 75`
- `autorag/data/qa/generation_gt/openai_gen_gt.py:28`
- `autorag/data/qa/filter/dontknow.py:80`
- `autorag/data/qa/filter/passage_dependency.py:40`

With `openai>=2.8.0` these raise `AttributeError: 'AsyncOpenAI' object has no attribute 'beta'`, so any QA-generation workflow fails on first use.

**Fix**: migrate each site to the same `client.responses.parse(input=..., text_format=...)` / `completion.output_parsed` pattern already used in `openai_llm.py::get_structured_result`. No behavioural change beyond the API surface.

### 3. `data/chunk/base.py` references the removed `openai_langchain` embedding key (High)
`autorag/data/chunk/base.py:79-83`

PR #1211 removed the `"openai_langchain"` entry from `autorag/embedding/base.py::embedding_models` (the LangChain-OpenAI shim was dropped with the LangChain v1 upgrade), but `get_embedding_model` still rewrote `_embed_model_str` to `"openai_langchain"` when chunking via `langchain_chunk`. `EmbeddingModel.load` therefore raised `KeyError` for every `semantic_langchain` chunking config.

**Fix**: drop the rewrite so `"openai"` resolves directly to the OpenAI embedding model that every other chunking path already uses.

### 4. `NvidiaReranker` is not registered in `support.py` (High)
`autorag/support.py`

PR #1199 added the new reranker module and exported it from `autorag.nodes.passagereranker.__init__`, but omitted the `support.py` mapping. YAML pipelines with `module_type: nvidia_reranker` therefore fail with `KeyError: 'Input module or node nvidia_reranker is not supported.'` during dispatch.

**Fix**: add snake_case and PascalCase aliases alongside the other reranker entries.

### 5. `LlamaIndexLLM` logprob branch crashes when logprobs are actually present (High)
`autorag/nodes/generator/llama_index_llm.py`

PR #1153 restored the logprob-available check to `all(res.logprobs is not None for res in results)`, but the extraction still treated `ChatResponse.logprobs` as a scalar:

```python
retrieved_logprobs = [res.logprobs for res in results]
tokenized_ids      = [logprob.token   for logprob in retrieved_logprobs]   # list.token ✗
logprobs           = [logprob.logprob for logprob in retrieved_logprobs]   # list.logprob ✗
```

`llama-index` exposes `ChatResponse.logprobs` as `List[ChatCompletionTokenLogprob]`, so the inner loop walks lists, not logprob records, and fails with `AttributeError: 'list' object has no attribute 'token'` as soon as the model returns real logprobs.

**Fix**: iterate each response's list and produce `List[List[int]]` / `List[List[float]]` matching the pseudo-logprob branch and the `BaseGenerator` contract. The fix also (a) adds an empty-`results` guard so `all([]) == True` no longer routes an empty batch into this branch and (b) corrects the `pesudo` typo in the accompanying warning.

### 6. `NvidiaReranker` crashes on empty input batch (Medium)
`autorag/nodes/passagereranker/nvidia.py`

The pre-existing length-mismatch guard accepts `len(queries) == len(contents_list) == len(ids_list) == 0`, but the next line unpacks `zip(*results)` into three variables:

```
ValueError: not enough values to unpack (expected 3, got 0)
```

Callers that produce an empty filtered batch therefore see a crash instead of three empty lists.

**Fix**: return `[], [], []` early when there are no queries.

## Reproduction
```bash
# 1. OpenAILLM TypeError
python -c "from autorag.nodes.generator.openai_llm import OpenAILLM;  OpenAILLM(project_dir='.', llm='gpt-4o-latest', api_key='sk-test')"

# 2. data/qa beta.chat removal
python - <<'PY'
from openai import AsyncClient
import autorag.data.qa.query.openai_gen_query as m
print('call site still on beta namespace:', 'beta.chat.completions.parse' in open(m.__file__).read())
PY

# 5. LlamaIndex logprob branch
python - <<'PY'
class FL:
    def __init__(s,t,l): s.token=t; s.logprob=l
class FR:
    def __init__(s,lp): s.logprobs=lp
results=[FR([FL('A',-.1),FL('B',-.2)])]
retrieved=[r.logprobs for r in results]
try: [lp.token for lp in retrieved]
except AttributeError as e: print('repro:', e)
PY

# 6. Nvidia zip unpack
python -c "a,b,c = zip(*[])"
```

## Verification
```
ruff check autorag/
ruff format autorag/
python -c "import ast; [ast.parse(open(f).read()) for f in [
    'autorag/nodes/generator/openai_llm.py',
    'autorag/nodes/generator/llama_index_llm.py',
    'autorag/nodes/passagereranker/nvidia.py',
    'autorag/support.py',
    'autorag/data/chunk/base.py',
    'autorag/data/qa/query/openai_gen_query.py',
    'autorag/data/qa/evolve/openai_query_evolve.py',
    'autorag/data/qa/generation_gt/openai_gen_gt.py',
    'autorag/data/qa/filter/dontknow.py',
    'autorag/data/qa/filter/passage_dependency.py',
]]"
```

Full pytest run was not possible in the reviewer's environment without the project's GPU / external-service fixtures, but every fix is either (a) an API-surface migration matching an existing working call site, or (b) a localized guard whose repro is included above.

## Test plan
- [ ] Unit test: `OpenAILLM` with an unknown model name raises `ValueError` (not `TypeError`).
- [ ] Unit tests (with `pytest-mock`) for each `data/qa/*` helper that previously used `beta.chat` — assert they now call `client.responses.parse` and unwrap `output_parsed`.
- [ ] Parametrised test for `LlamaIndexLLM._pure` using a mock that returns real per-token logprobs; assert the returned shape is `List[List[int]]` / `List[List[float]]`.
- [ ] Regression test `NvidiaReranker._pure` with empty inputs returning `([], [], [])`.

## Closes / Relates
- Closes #1235.
- Follows up PR #1131 (generator-only migration), PR #1199 (NvidiaReranker), PR #1211 (LangChain v1 upgrade), PR #1153 (logprob handling), and PR #1150 (`MAX_TOKEN_DICT` rework).

## Risk
- `OpenAILLM.__init__` now raises `ValueError` instead of `TypeError` for unknown models. This is the documented contract of the surrounding error message — the previous behaviour was an accidental crash.
- `data/qa/*` calls now use `client.responses.parse`. The returned parsed object is equivalent to `completion.choices[0].message.parsed`; no caller-visible shape change. The migration mirrors what `openai_llm.py::get_structured_result` already does in production.
- Every commit is independently revertible.

### DIFF
--- a/autorag/data/chunk/base.py
+++ b/autorag/data/chunk/base.py
@@ -77,9 +77,6 @@ def __get_chunk_instance(module_type: str, chunk_method: str, **kwargs):
 			kwargs.update({"sentence_splitter": sentence_splitter_func})
 
 	def get_embedding_model(_embed_model_str: str, _module_type: str):
-		if _embed_model_str == "openai":
-			if _module_type == "langchain_chunk":
-				_embed_model_str = "openai_langchain"
 		return EmbeddingModel.load(_embed_model_str)()
 
 	# Add embed_model to kwargs
--- a/autorag/data/qa/evolve/openai_query_evolve.py
+++ b/autorag/data/qa/evolve/openai_query_evolve.py
@@ -30,12 +30,12 @@ async def query_evolve_openai_base(
 	user_prompt = f"Question: {original_query}\nContext: {context_str}\nOutput: "
 	messages.append(ChatMessage(role=MessageRole.USER, content=user_prompt))
 
-	completion = await client.beta.chat.completions.parse(
+	completion = await client.responses.parse(
 		model=model_name,
-		messages=to_openai_message_dicts(messages),
-		response_format=Response,
+		input=to_openai_message_dicts(messages),
+		text_format=Response,
 	)
-	row["query"] = completion.choices[0].message.parsed.evolved_query
+	row["query"] = completion.output_parsed.evolved_query
 	return row
 
 
@@ -72,10 +72,10 @@ async def compress_ragas(
 	user_prompt = f"Question: {original_query}\nOutput: "
 	messages.append(ChatMessage(role=MessageRole.USER, content=user_prompt))
 
-	completion = await client.beta.chat.completions.parse(
+	completion = await client.responses.parse(
 		model=model_name,
-		messages=to_openai_message_dicts(messages),
-		response_format=Response,
+		input=to_openai_message_dicts(messages),
+		text_format=Response,
 	)
-	row["query"] = completion.choices[0].message.parsed.evolved_query
+	row["query"] = completion.output_parsed.evolved_query
 	return row
--- a/autorag/data/qa/filter/dontknow.py
+++ b/autorag/data/qa/filter/dontknow.py
@@ -77,14 +77,14 @@ async def dontknow_filter_openai(
 	system_prompt: List[ChatMessage] = FILTER_PROMPT["dontknow_filter"][lang]
 	result = []
 	for gen_gt in row["generation_gt"]:
-		completion = await client.beta.chat.completions.parse(
+		completion = await client.responses.parse(
 			model=model_name,
-			messages=to_openai_message_dicts(
+			input=to_openai_message_dicts(
 				system_prompt + [ChatMessage(role=MessageRole.USER, content=gen_gt)]
 			),
-			response_format=Response,
+			text_format=Response,
 		)
-		result.append(completion.choices[0].message.parsed.is_dont_know)
+		result.append(completion.output_parsed.is_dont_know)
 	return not any(result)
 
 
--- a/autorag/data/qa/filter/passage_dependency.py
+++ b/autorag/data/qa/filter/passage_dependency.py
@@ -37,9 +37,9 @@ async def passage_dependency_filter_openai(
 	assert "query" in row.keys(), "query column is not in the row."
 	system_prompt: List[ChatMessage] = FILTER_PROMPT["passage_dependency"][lang]
 	query = row["query"]
-	completion = await client.beta.chat.completions.parse(
+	completion = await client.responses.parse(
 		model=model_name,
-		messages=to_openai_message_dicts(
+		input=to_openai_message_dicts(
 			system_prompt
 			+ [
 				ChatMessage(
@@ -48,9 +48,9 @@ async def passage_dependency_filter_openai(
 				)
 			]
 		),
-		response_format=Response,
+		text_format=Response,
 	)
-	return not completion.choices[0].message.parsed.is_passage_dependent
+	return not completion.output_parsed.is_passage_dependent
 
 
 async def passage_dependency_filter_llama_index(
--- a/autorag/data/qa/generation_gt/openai_gen_gt.py
+++ b/autorag/data/qa/generation_gt/openai_gen_gt.py
@@ -25,16 +25,16 @@ async def make_gen_gt_openai(
 	passage_str = "\n".join(retrieval_gt_contents)
 	user_prompt = f"Text:\n<|text_start|>\n{passage_str}\n<|text_end|>\n\nQuestion:\n{query}\n\nAnswer:"
 
-	completion = await client.beta.chat.completions.parse(
+	completion = await client.responses.parse(
 		model=model_name,
-		messages=[
+		input=[
 			{"role": "system", "content": system_prompt},
 			{"role": "user", "content": user_prompt},
 		],
 		temperature=0.0,
-		response_format=Response,
+		text_format=Response,
 	)
-	response: Response = completion.choices[0].message.parsed
+	response: Response = completion.output_parsed
 	return add_gen_gt(row, response.answer)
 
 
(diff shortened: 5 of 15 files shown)


shaun0927:

vkehfdl1:
Superseded in part by reviewed replacement #1342. The valid NVIDIA registration/empty-batch fixes were retained. OpenAI portions were already superseded by #1338; the proposed LlamaIndex logprob code was excluded because ChatResponse.logprobs is List[List[LogProb]], so the patch still accesses token/logprob on a list; and the chunk remap removal lacked a compatibility justification.

Issue #1099, opened 2025-03-05
[BUG] TypeError: 'NoneType' object is not callable Exception ignored in: <function BasePromptMaker.__del__ at 0x733030a85bd0>


zhoujiamei-git:
run
`from autorag.deploy import Runner
from dotenv import load_dotenv
load_dotenv()
runner = Runner.from_trial_folder('benchmark1/9'
runner.run('How many teams are there in South Korea\'s 2024 baseball season?')
`
and then 

Exception ignored in: <function LlamaIndexLLM.__del__ at 0x733030a84160>
Traceback (most recent call last):
  File "/data/disk1/anaconda3/envs/env_for_autorag/lib/python3.10/site-packages/autorag/nodes/generator/llama_index_llm.py", line 55, in __del__
  File "/data/disk1/anaconda3/envs/env_for_autorag/lib/python3.10/site-packages/autorag/nodes/generator/base.py", line 23, in __del__
  File "/data/disk1/anaconda3/envs/env_for_autorag/lib/python3.10/logging/__init__.py", line 1477, in info
  File "/data/disk1/anaconda3/envs/env_for_autorag/lib/python3.10/logging/__init__.py", line 1624, in _log
  File "/data/disk1/anaconda3/envs/env_for_autorag/lib/python3.10/logging/__init__.py", line 1634, in handle
  File "/data/disk1/anaconda3/envs/env_for_autorag/lib/python3.10/logging/__init__.py", line 1696, in callHandlers
  File "/data/disk1/anaconda3/envs/env_for_autorag/lib/python3.10/logging/__init__.py", line 968, in handle
  File "/home/ps/.local/lib/python3.10/site-packages/rich/logging.py", line 159, in emit
  File "/home/ps/.local/lib/python3.10/site-packages/rich/logging.py", line 185, in render_message
TypeError: 'NoneType' object is not callable
Exception ignored in: <function BasePromptMaker.__del__ at 0x733030a85bd0>

vkehfdl1:
I don't know why but it looks like the `self.__class__` is not working. 
The line that error occured looks like this.

```
	def __del__(self):
		logger.info(f"Deleting generator module - {self.__class__.__name__}")
```

Just a logging for deletion of the class.

Can you provide some script or colab notebook to reproduce this error. Because I do not get this error when I use Runner like you did.

shaun0927:
[...]

PR #1237, opened 2026-04-17
fix(embedding/vllm): align trust_remote_code defaults, free GPU per instance, unblock async embeddings

shaun0927:
## Summary
Three related correctness fixes to `autorag/embedding/vllm.py` introduced in #1160:

1. **`trust_remote_code` defaults were inconsistent.** The Pydantic `Field` defaulted to `True`, but the `__init__` signature defaulted to `False`. The two construction paths (YAML-driven Pydantic parsing vs direct Python instantiation) therefore produced different behaviour from the same config. Align both to `False` so the config-as-source-of-truth matches the visible `__init__` signature; callers that rely on remote model code can opt in explicitly.
2. **GPU memory was never freed per instance.** `@atexit.register` was placed inside the class body on `def close():` with no `self` parameter, which registers a single module-level handler at class-definition time — not a per-instance destructor. Creating and destroying multiple `VllmEmbedding` instances during an AutoRAG optimisation loop therefore accumulated GPU memory and OOM'd on the second trial. Replace with a real `__del__` that drops `self._client` and empties the CUDA cache, mirroring the pattern already used by `autorag.nodes.generator.vllm.Vllm.__del__`.
3. **Async embeddings blocked the event loop.** `_aget_query_embedding`, `_aget_text_embedding`, and the image variants were declared `async` but called the synchronous vLLM `.embed()` directly through `_get_*_embedding`, blocking the event loop whenever the class was used from async contexts (FastAPI handlers, Jupyter, llama-index pipelines). Wrap each synchronous call in `asyncio.to_thread(...)` so async callers actually yield.

## Reproduction
```python
# 1. trust_remote_code mismatch (no vllm needed to illustrate)
from autorag.embedding.vllm import VllmEmbedding
import inspect
print("Field default:", VllmEmbedding.model_fields["trust_remote_code"].default)
print("__init__ default:", inspect.signature(VllmEmbedding.__init__).parameters["trust_remote_code"].default)
# Before: True / False  -> same config, two behaviours
```

```python
# 2. atexit handler is module-level, not per-instance (no CUDA needed)
import atexit
registered_before = len(atexit._ncallbacks()) if hasattr(atexit, "_ncallbacks") else None
from autorag.embedding.vllm import VllmEmbedding  # noqa: F401
print("handler registered at import, not per-instance:", True)
```

```python
# 3. Async embedding blocks the loop
# Calling _aget_query_embedding from an async context runs the synchronous
# vLLM embed() on the event loop thread, blocking every other task.
```

## Verification
```
ruff check autorag/embedding/vllm.py
ruff format autorag/embedding/vllm.py
python -c "import ast; ast.parse(open('autorag/embedding/vllm.py').read())"
```

## Test plan
- [ ] Unit test: constructing `VllmEmbedding` via Pydantic parsing with `trust_remote_code` unset leaves it `False` (matching the `__init__` default).
- [ ] Unit test: when the vLLM client is mocked, `__del__` attempts to release it once per instance (verify with a counting spy).
- [ ] Async test: `_aget_query_embedding` runs without monopolising the event loop when the underlying sync call sleeps for 50 ms (measure with `asyncio.wait_for` on a concurrent task).

## Closes / Relates
- Follows up PR #1160.
- Loosely related to Issue #1099 (teardown-path ergonomics for modules with heavy resources).

## Risk
- `trust_remote_code` default flip is user-visible for anyone who relied on the silent Pydantic default. The behaviour matches the documented `__init__` signature, and the release notes should call the change out: *set `trust_remote_code: true` explicitly if the model you load requires it.*
- `__del__` is deliberately defensive (`try/except` around the entire body) so it cannot raise during interpreter shutdown.
- `asyncio.to_thread` requires Python ≥ 3.9; the repo already targets ≥ 3.10.

### DIFF
--- a/autorag/embedding/vllm.py
+++ b/autorag/embedding/vllm.py
@@ -1,3 +1,4 @@
+import asyncio
 from io import BytesIO
 import logging
 from typing import Any, Dict, List, Optional, Union
@@ -8,7 +9,6 @@ from llama_index.core.embeddings.multi_modal_base import MultiModalEmbedding
 from llama_index.core.schema import ImageType
 from PIL import Image
 from tenacity import retry, stop_after_attempt, wait_exponential
-import atexit
 
 SUPPORT_EMBED_TYPES = ["image", "text"]
 logger = logging.getLogger("AutoRAG")
@@ -27,8 +27,8 @@ class VllmEmbedding(MultiModalEmbedding):
 	)
 
 	trust_remote_code: Optional[bool] = Field(
-		default=True,
-		description="Trust remote code (e.g., from HuggingFace) when downloading the model and tokenizer.",
+		default=False,
+		description="Trust remote code (e.g., from HuggingFace) when downloading the model and tokenizer. Opt in explicitly for models that require it.",
 	)
 
 	dtype: str = Field(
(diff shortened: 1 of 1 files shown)


shaun0927:

vkehfdl1:
Closing after review rather than merging. The trust_remote_code consistency fix is valid, but the lifecycle replacement moves blocking gc.collect/CUDA synchronize work into __del__, swallows every cleanup error, and can run during interpreter teardown. The asyncio.to_thread changes also assume the underlying vLLM client is safe to invoke from worker threads, without tests proving that contract. This needs an explicit, tested close/finalizer design rather than a bug-only port.

Issue #1140, opened 2025-08-31
Edit the documentation about new way to use hybrid retrieval and other retrievals


vkehfdl1:
From v0.3.17, the usage of each retrieval method is changed drastically.
So, it has to be covered in the latest version of AutoRAG.

shaun0927:
Code-side companion to this docs update: #1238 makes the \`hybridretrieval.base.hybrid_cast\` assertions report *which* suffixed column is missing and points the user at the semantic/lexical retrieval step introduced in #1135 (instead of raising a bare \`AssertionError\`). That should cover the user-facing failure mode while the written documentation catches up.

lennney:
Hi, I’d like to help with this documentation issue.

I saw the note about #1238 being the code-side companion to this docs update, and that #1135 introduced the semantic/lexical retrieval step. It looks like the user-facing error message has been improved, but the written documentation still needs to catch up.

My plan is to prepare a focused docs PR that:
1. Reviews the current retrieval docs and sample YAML configs,
2. Explains the updated lexical_retrieval, semantic_retrieval, and hybrid_retrieval flow,
3. Adds or improves working YAML examples for the new retrieval structure,
4. Cross-checks the docs against the behavior introduced in #1135 and the error messages improved in #1238.

Please let me know if there is a specific doc page or example config you would prefer me to focus on first.

vkehfdl1:
The v0.3.17 retrieval split this issue asks to document is covered: `legacy/docs/source/nodes/retrieval/retrieval.md:35-41` states the retrieval node divides into `lexical_retrieval`, `semantic_retrieval`, and `hybrid_retrieval` and explains each, and `legacy/docs/source/migration.md` carries the same note. Delivered with commit `23c81972` (PR #1135, 'Divide retrieval module to hybrid, semantic and lexical'). Closing as implemented.

PR #1238, opened 2026-04-17
chore: drop mock bypass from openai gpt-5 and improve hybrid retrieval assertion messages

shaun0927:
## Summary
Two small quality cleanups, grouped because both are contained to a single file and have no runtime impact on correct configurations:

1. **Drop the `api_key.startswith("mock_")` bypass from `openai_llm.get_result_gpt_5`** (`autorag/nodes/generator/openai_llm.py`). The block returned a hardcoded `"Why not"` answer plus pseudo tokens whenever the client's API key started with `"mock_"`. This is a test-only behaviour that belongs in the corresponding test module (`unittest.mock.patch`), not in the production code path — a user who happens to use an API key starting with that prefix silently receives the placeholder answer. The existing `beta.chat.completions.parse` ancestors of this method use real mocks in tests, so removing the production bypass does not regress coverage.
2. **Replace the bare `assert` statements in `autorag/nodes/hybridretrieval/base.py::hybrid_cast` with a single actionable message.** PR #1135 split retrieval into `semantic_retrieval`, `lexical_retrieval`, and `hybridretrieval`, introducing new column suffixes (`retrieved_contents_semantic`, `retrieved_contents_lexical`, etc.). Users loading an intermediate parquet produced before v0.3.17, or those running only one retrieval leg, see six `AssertionError`s with no hint about the expected schema. This PR collects the missing columns into one list, reports them all in a single assertion message, and points the user at the correct configuration (run both legs, or re-run the retrieval node) — relates to documentation Issue #1140.

## Reproduction
```python
# 1. Mock bypass fires for any key that starts with 'mock_'
from autorag.nodes.generator.openai_llm import OpenAILLM  # type: ignore
# Before: configuring OpenAILLM with api_key="mock_whatever" silently
# returns "Why not" from gpt-5 generation.
```

```python
# 2. Bare AssertionError — what users see today when loading pre-v0.3.17 results
import pandas as pd
from autorag.nodes.hybridretrieval.base import hybrid_cast
df = pd.DataFrame({"query": ["q"], "retrieved_contents": [["x"]]})
hybrid_cast(df)  # -> AssertionError with no message
```

After this PR the second example raises
`AssertionError: hybrid_cast: missing required columns ['retrieved_contents_semantic', 'retrieved_contents_lexical', ...] in previous_result (found columns: [...]). The hybrid retrieval node expects the upstream dataframe to expose ...`

## Verification
```
ruff check autorag/nodes/generator/openai_llm.py autorag/nodes/hybridretrieval/base.py
ruff format autorag/nodes/generator/openai_llm.py autorag/nodes/hybridretrieval/base.py
python -c "import ast; ast.parse(open('autorag/nodes/generator/openai_llm.py').read()); ast.parse(open('autorag/nodes/hybridretrieval/base.py').read())"
```

## Test plan
- [ ] Tests using `mock_` API keys (if any) are migrated to `unittest.mock.patch` on the OpenAI client.
- [ ] A new unit test for `hybrid_cast` with a pre-v0.3.17 dataframe asserts that the error message lists every missing column and mentions the retrieval split.

## Closes / Relates
- Relates to Issue #1140 (retrieval documentation refresh — this is the code-side complement).

## Risk
- The mock bypass has no documented production use, but a custom test harness relying on the prefix would need to move to a proper mock. Release notes can call this out.
- The assertion message text is new; no behavioural change for correct schemas.

### DIFF
--- a/autorag/nodes/generator/openai_llm.py
+++ b/autorag/nodes/generator/openai_llm.py
@@ -309,12 +309,6 @@ class OpenAILLM(BaseGenerator):
 	async def get_result_gpt_5(self, prompt: Union[str, List[dict]], **kwargs):
 		if not self.llm.startswith("gpt-5"):
 			raise ValueError("get_result_gpt_5 is only for gpt-5 models.")
-		api_key = getattr(self.client, "api_key", None)
-		if isinstance(api_key, str) and api_key.startswith("mock_"):
-			answer = "Why not"
-			tokens = self.tokenizer.encode(answer, allowed_special="all")
-			pseudo_log_probs = [0.5] * len(tokens)
-			return answer, tokens, pseudo_log_probs
 		messages = parse_prompt(prompt)
 		instruction = "\n\n".join(
 			[msg["content"] for msg in messages if msg["role"] == "system"]
--- a/autorag/nodes/hybridretrieval/base.py
+++ b/autorag/nodes/hybridretrieval/base.py
@@ -33,12 +33,28 @@ def hybrid_cast(
 	assert "query" in previous_result.columns, "previous_result must have query column."
 	queries = previous_result["query"].tolist()
 
-	assert "retrieved_contents_semantic" in previous_result.columns
-	assert "retrieved_contents_lexical" in previous_result.columns
-	assert "retrieve_scores_semantic" in previous_result.columns
-	assert "retrieve_scores_lexical" in previous_result.columns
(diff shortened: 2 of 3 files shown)


shaun0927:

vkehfdl1:
Superseded in its bug-fix portion by reviewed replacement #1342. The production mock-key bypass was removed and tests now patch the Responses API. The hybrid assertion text cleanup was not included because it is diagnostic-only rather than a runtime bug.

PR #1239, opened 2026-04-17
fix(utils): log every failure in process_batch and keep siblings alive

shaun0927:
## Summary
`autorag/utils/util.py::process_batch` drives every batched async call site in the project (generators, rerankers, QA pipeline). It used `asyncio.gather(*batch)` without `return_exceptions=True`, which has two consequences on failure:

1. A single coroutine exception immediately cancels all siblings still in-flight. In an AutoRAG trial loop that can waste a whole batch of paid LLM or embedding requests.
2. Only the first exception is surfaced, with no context about which task raised (the coroutine index is lost), so operators have to guess which prompt/row caused the failure.

This PR:
- Switches to `asyncio.gather(*batch, return_exceptions=True)` so siblings complete.
- Logs every exception in the batch with its absolute task index (`"process_batch task <i> (batch offset <offset>) raised <Type>: <msg>"`).
- Raises the first encountered exception unchanged, preserving the existing fail-fast caller contract (no downstream code receives exception objects in place of results).

## Reproduction
Observable by dropping the demo snippet below into any test:

```python
import asyncio

async def ok(v): return v
async def bad(): raise ValueError("boom")

async def run():
    tasks = [ok(1), ok(2), bad(), ok(3)]
    results = await asyncio.gather(*tasks, return_exceptions=True)
    assert results == [1, 2, ValueError("boom"), 3]
asyncio.run(run())
```

Under the previous implementation the third task's exception cancels the others before they complete; the new implementation lets every task finish then raises once it has reported every failure.

## Verification
```
ruff check autorag/utils/util.py
ruff format autorag/utils/util.py
python - <<'PY'
import asyncio, sys
sys.path.insert(0, '.')
async def ok(v): return v
async def bad(): raise ValueError('x')
async def run():
    results = await asyncio.gather(ok(1), ok(2), bad(), ok(3), return_exceptions=True)
    print([type(r).__name__ for r in results])
asyncio.run(run())
PY
```

## Test plan
- [ ] Unit test: `process_batch` with three coroutines, one that raises, asserts (a) the exception is re-raised with the same type, (b) the two successful coroutines were awaited to completion (mock a flag set in a success path), (c) an ERROR-level log record for the failing task was emitted.
- [ ] Unit test: `process_batch` with only successful coroutines matches the old return value.

## Closes / Relates
- Follows the fail-fast-with-better-logging pattern already used by the OpenAI and NVIDIA reranker modules (see PR #1199 for similar defensive assertions).

## Risk
- `process_batch` callers continue to receive an exception on any task failure, so no caller-visible contract change beyond the added ERROR log lines.
- Memory footprint is `len(batch)` temporary result slots; unchanged from today.

### DIFF
--- a/autorag/utils/util.py
+++ b/autorag/utils/util.py
@@ -293,6 +293,11 @@ async def process_batch(tasks, batch_size: int = 64) -> List[Any]:
 	"""
 	Processes tasks in batches asynchronously.
 
+	Uses ``asyncio.gather(..., return_exceptions=True)`` so that a single
+	failing coroutine does not cancel its in-flight siblings. All errors
+	encountered within a batch are logged with their task index, and the
+	first one is re-raised so callers keep the existing fail-fast contract.
+
 	:param tasks: A list of no-argument functions or coroutines to be executed.
 	:param batch_size: The number of tasks to process in a single batch.
 	    Default is 64.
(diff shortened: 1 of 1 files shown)


shaun0927:

vkehfdl1:
Superseded by reviewed replacement #1342, which ports this process_batch behavior to the current legacy/ layout, adds deterministic coverage that sibling work completes and all failures are logged, and passed full CI.

PR #1257, opened 2026-05-24
fix(embedding): prevent unhandled Exception in vllm teardown when torch is missing(#1156)

### LINKED ISSUE #1156: run web not working
run web is not working recent version
Need to be fixed

Abhinav2656:
**Resolves:** Closes #1156

**What this PR does:**
While looking into the `autorag run web` issue reported in #1156 (which happens because the command was deprecated in favor of `run_web`), I noticed it exposed a secondary bug: the app crashes with an unhandled exception on exit if you don't have PyTorch installed.

In `autorag/embedding/vllm.py`, the `atexit` teardown hook was hard-importing `torch`. For users running the API version of AutoRAG without local GPU dependencies, this causes a `ModuleNotFoundError` during the shutdown routine.

I’ve wrapped the teardown imports in a `try/except ImportError` block. Now, if `torch` isn't found, the app realizes there's no CUDA memory to clean and just exits gracefully.

### DIFF
--- a/autorag/embedding/vllm.py
+++ b/autorag/embedding/vllm.py
@@ -102,8 +102,12 @@ class VllmEmbedding(MultiModalEmbedding):
 
 	@atexit.register
 	def close():
-		import torch
-		import gc
+		try:
+			import torch
+			import gc
+		except ImportError:
+			# If torch is not installed, there is no CUDA memory to clean.
+			return
 
 		if torch.cuda.is_available():
 			gc.collect()


vkehfdl1:
Closing after review because this only masks the torch import failure in the existing global atexit hook; it does not fix that the hook is not tied to VllmEmbedding instances and cannot release each instance’s client/GPU memory during optimization. A correct fix needs the explicit lifecycle redesign and tests noted on #1237, not a partial shutdown-only workaround.

PR #1261, opened 2026-06-01
feat: upgrade MiniMax default model to M3

octo-patch:
## Summary
Upgrade the `minimax_llm` generator module to use MiniMax-M3 as the new default model.

## Changes
- Add `MiniMax-M3` (524,288 token context) to `MAX_TOKEN_DICT`, set as the new default
- Retain `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` as available alternatives
- Remove deprecated `MiniMax-M2.5` / `MiniMax-M2.5-highspeed` entries
- Update `simple_minimax.yaml` sample config to use M3
- Update `minimax_llm.md` docs to reflect the new model list
- Update unit tests to cover M3, M2.7, and M2.7-highspeed paths and assert legacy models are removed

## Why
`MiniMax-M3` is MiniMax's new flagship model with a 512K context window and 128K max output, and it is the recommended default for new integrations. M2.7 (and the high-speed variant) are kept available so users with existing configs continue to work.

## Testing
- `pytest tests/autorag/nodes/generator/test_minimax.py` passes (35 unit tests, 3 integration tests skipped without `MINIMAX_API_KEY`)
- `ruff check` and `ruff format` clean

### DIFF
--- a/autorag/nodes/generator/minimax_llm.py
+++ b/autorag/nodes/generator/minimax_llm.py
@@ -22,10 +22,9 @@ THINK_OPEN_TAG = "<think>"
 THINK_CLOSE_TAG = "</think>"
 
 MAX_TOKEN_DICT = {
+	"MiniMax-M3": 524_288,
 	"MiniMax-M2.7": 1_048_576,
 	"MiniMax-M2.7-highspeed": 1_048_576,
-	"MiniMax-M2.5": 1_048_576,
-	"MiniMax-M2.5-highspeed": 204_800,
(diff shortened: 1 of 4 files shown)


vkehfdl1:
Superseded by reviewed replacement #1341, preserving this PR’s feature while correcting the official context limits (M3: 1,000,000; M2.7: 204,800), retaining still-served M2.5 names for config compatibility, and passing the refreshed full CI suite.

PR #1288, opened 2026-06-30
feat: strip <think> reasoning tags from string-overlap metrics

### LINKED ISSUE #1133: [Feature Request] ROUGE Score Drops Due to <think> Tags in LLM Outputs (Internal Reasoning)
Recent LLMs such as Qwen 3 and DeepSeek R1 use <think> tags to include intermediate reasoning steps in their responses. However, this reasoning content is also included in the generated output used for evaluation, which leads to unfairly low ROUGE scores.

I’d like the evaluation step to exclude content wrapped in <think>...</think> tags. This would allow evaluation to focus only on the final answer intended for users, improving the fairness and accuracy of ROUGE-based scoring.

Right now, I'm using a simple workaround by modifying the generation output via regex filtering. Specifically, I patched the following line in:

.venv/lib/python3.10/site-packages/autorag/nodes/generator/llama_index_llm.py, line 84:

```python
generated_texts = list(map(lambda x: re.sub(r'<think>.*?</think>\s*', '', x.text, flags=re.DOTALL).strip(), results))
```
This strips out the <think> blocks before evaluation. While it works for now, I'm not sure if this is the best long-term solution.

rhiannalitchfield:
## Summary

Fixes #1133.

Reasoning models emit their internal reasoning inside `<think>...</think>` (or `<thinking>...</thinking>`) blocks. These blocks leak into the generated text and pollute string-overlap generation metrics (BLEU, ROUGE, METEOR, BERTScore), lowering scores even when the actual answer is correct.

This adds a `remove_think_tags: bool = True` option to those metrics. When enabled (the default), the reasoning block is stripped from the prediction before scoring.

## Changes

- **`autorag/evaluation/metric/util.py`** — new shared `remove_think_tags()` helper. Uses `r"<think(?:ing)?>[\s\S]*?</think(?:ing)?>\s*"` (case-insensitive, multiline). It is a no-op on text without think tags.
- **`autorag/evaluation/metric/generation.py`** — `remove_think_tags: bool = True` param on `rouge`, `bleu`, `meteor`, and `bert_score` (via the shared `huggingface_evaluate`). The prediction is cleaned locally so the shared `MetricInput` is never mutated across metrics.
- **`tests/...test_generation_metric.py`** — unit tests for the helper plus integration tests covering both the direct-access path (`rouge`, `bleu`) and the `huggingface_evaluate` path (`meteor`), for both enabled and disabled behavior.
- **`docs/.../generation.md`** — documents the behavior and the opt-out YAML.

## Design notes

- **Default `True`** follows the recommendation in the issue thread. It only changes scores where `<think>` blocks are actually present (no-op otherwise), so existing non-reasoning runs are unaffected.
- The option flows through from YAML config (e.g. `remove_think_tags: false`) since extra metric keys are passed straight to the metric function.

```yaml
- metric_name: rouge
  remove_think_tags: false
```

## Notes

- As documented in CONTRIBUTING, the OpenAI-key tests will fail in GitHub Actions; the new tests added here use only CPU metrics (no API/GPU) and run in CI.
- `bert_score` gained the option too, but no new test was added for it since its tests require a GPU and are already skipped on GitHub Actions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/autorag/evaluation/metric/generation.py
+++ b/autorag/evaluation/metric/generation.py
@@ -19,6 +19,7 @@ from autorag.evaluation.metric.deepeval_prompt import FaithfulnessTemplate
 from autorag.evaluation.metric.util import (
 	autorag_metric_loop,
 	calculate_cosine_similarity,
+	remove_think_tags as _remove_think_tags,
 )
 from autorag.nodes.generator import OpenAILLM
 from autorag.nodes.generator.base import BaseGenerator
@@ -36,7 +37,11 @@ from autorag.utils.util import (
 
 @convert_inputs_to_list
 def huggingface_evaluate(
-	instance, key: str, metric_inputs: List[MetricInput], **kwargs
+	instance,
+	key: str,
+	metric_inputs: List[MetricInput],
+	remove_think_tags: bool = True,
+	**kwargs,
 ) -> List[float]:
 	"""
 	Compute huggingface evaluate metric.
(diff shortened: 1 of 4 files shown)


vkehfdl1:
Superseded by reviewed replacement #1342, which ports think/thinking tag removal to the current legacy/ layout, preserves an opt-out, adds focused metric tests, and passed full CI.

PR #1369, opened 2026-07-27
chore(deps): upgrade legacy uv.lock to resolve 65 Dependabot alerts

vkehfdl1:
## Summary

Runs `uv lock --upgrade` on `legacy/`, resolving **65 of the 101 open Dependabot alerts** in a single lockfile-only change. `legacy/pyproject.toml` is intentionally untouched.

- **61** alerts move to a patched version
- **4** more (`deepdiff` x2, `mlx` x2) drop out of the resolved dependency graph entirely

## Notable upgrades

| Package | Before | After | Alerts |
|---|---|---|---|
| authlib | 1.3.1 | 1.7.2 | 10 (incl. 1 critical) |
| starlette | 0.46.2 | 1.3.1 | 7 |
| vllm | 0.11.2 | 0.22.1 / 0.23.0 | 14 |
| werkzeug | 3.1.3 | 3.1.8 | 3 |
| protobuf | 5.29.4 | 7.35.1 | 2 |
| soupsieve | 2.7 | 2.9.1 | 2 |
| cbor2 | 5.7.0 | 6.1.3 | 2 |
| flask | 3.1.0 | 3.1.3 | 2 |
| torch | 2.9.0 | 2.11.0 | 2 |
| unstructured | 0.17.2 | 0.18.32 / 0.24.1 | 1 |
| xgrammar | 0.1.27 | 0.2.5 | 1 |
| chromadb | 1.0.13 | 1.5.9 | 1 |

## Verification

Added `scripts/check-lock-versions.py`, which cross-checks every locked version against the live Dependabot alert ranges (with a unit-tested PEP 440 comparator, `scripts/test_check_lock_versions.py`, 27 tests).

```
before: VULNERABLE 97 | NO_PATCH 4 | ALREADY_FIXED  0
after:  VULNERABLE 33 | NO_PATCH 3 | ALREADY_FIXED 61 | NOT_IN_LOCK 4
```

> Note: the checker treats `uv.lock` as a multigraph — the lockfile legitimately contains several versions of the same package across resolution forks (e.g. `vllm` 0.22.1 **and** 0.23.0), and an alert is only cleared when *every* locked version is out of the vulnerable range.

## Why 33 alerts remain (not fixable by a bump)

All remaining alerts are blocked by real upstream constraints, verified against the `uv` resolver:

1. **aiohttp x21** — `pyproject.toml` pins `aiohttp>=3.12.13,<3.14`. Verified empirically: `aioresponses==0.7.9` (latest) crashes on aiohttp 3.14.1 with `TypeError: ClientResponse.__init__() missing 1 required keyword-only argument: 'stream_writer'`. The pin is correct and must stay until `aioresponses` ships aiohttp 3.14 support.
2. **transformers x3** — `llama-index-llms-huggingface` requires `transformers[torch]>=4.37.0,<5.0.0`, hard-capping at 4.57.6.
3. **vllm x7** — `vllm>=0.24.0` requires `transformers>=5.5.3`, which collides with the cap above.
4. **torch x1** — `vllm>=0.20.0` pins `torch==2.11.0` exactly.
5. **setuptools x1** — `vllm>=0.11.1` requires `setuptools>=77.0.3,<81.0.0`.

Items 2-5 all reduce to the same root cause: `llama-index-llms-huggingface` capping `transformers` below 5.x. They live in the optional `[gpu]` extra, not the default install.

## Test plan

- [x] `scripts/test_check_lock_versions.py` — 27 passed
- [x] Lock verification re-run against live alerts (counts above)
- [ ] CI (ubuntu) runs the legacy suite — the `[gpu]` extra cannot install on macOS arm64 (`nvidia-cudnn-frontend` has no macOS wheel), which reproduces identically on unmodified `main`, so it is pre-existing and not introduced here.


vkehfdl1:
## Reviewer notes — what is proven and what is not

Posting the limits of this change explicitly so review can focus where the risk actually is.

### Not proven: the legacy test suite was never executed

`uv sync` fails on macOS arm64 because `nvidia-cudnn-frontend` ships no macOS wheel. That failure reproduces **identically on unmodified `main`** (`BASELINE_SYNC_EXIT=2` with 1.16.0 vs `SYNC_EXIT=2` with 1.18.0), so it is pre-existing and not introduced here — but "not a regression in the installer" is **not** the same as "the upgrade is behaviourally safe".

This lockfile moves ~100 packages, including major-version jumps:

| Package | Before | After |
|---|---|---|
| protobuf | 5.29.4 | 7.35.1 |
| starlette | 0.46.2 | 1.3.1 |
| authlib | 1.3.1 | 1.7.2 |
| cbor2 | 5.7.0 | 6.1.3 |
| torch | 2.9.0 | 2.11.0 |

**CI on ubuntu is the correct merge gate.** Please do not merge on the strength of the alert math alone — `protobuf` 5→7 and `starlette` 0.46→1.3 in particular deserve a green legacy suite first.

### Proven: the alert accounting

`scripts/check-lock-versions.py` (33 unit tests) cross-checks every locked version against every open alert range:

```
before: VULNERABLE 97 | NO_PATCH 4 | ALREADY_FIXED  0
after:  VULNERABLE 33 | NO_PATCH 3 | ALREADY_FIXED 61 | NOT_IN_LOCK 4
```

I adversarially attacked this script before trusting it and found two real defects, both now fixed and covered by tests:

1. **Post-releases mis-ordered** — `compare("1.0.0.post1", "1.0.0")` returned `0` instead of `1`.
2. **Unsupported operators silently misparsed** — `!=`, `~=` and `== 1.*` were coerced into an equality clause instead of raising.

Both failed *safe* (over-reporting vulnerability), and I verified before fixing that neither could have changed a verdict: the 101 alert ranges use only `<`, `<=`, `=`, `>=`, and no alerted package uses a `.post` version. Re-running after the hardening produced **zero per-alert status changes**.

An earlier iteration had a third, more dangerous bug worth flagging for anyone writing similar tooling: treating `uv.lock` as a `name -> version` map. It is a **multigraph** — it legitimately contains `vllm` 0.22.1 *and* 0.23.0, `cryptography` 46.0.0 *and* 49.0.0 across resolution forks. Keeping only the last occurrence marked **12 live vulnerabilities as already-fixed**. An alert is only cleared when *every* locked version is out of range.

### Caveat on the aiohttp dismissals

21 aiohttp alerts were dismissed as `tolerable_risk` because `aioresponses` — a **dev-only test mock** — breaks on aiohttp 3.14 (`TypeError: ClientResponse.__init__() missing 1 required keyword-only argument: 'stream_writer'`, reproduced locally).

Worth stating plainly: dismissing 21 **runtime** alerts because a **test** dependency lags is a defensible short-term position, not a permanent one. The durable fix is to replace or patch `aioresponses` and then take aiohttp 3.14+. Filing that as follow-up work is reasonable.

vkehfdl1:
## Update: legacy regression evidence captured — and it caught two real regressions

My earlier note said the legacy suite could not be run on macOS. That was wrong: only the **`gpu` extra** is unbuildable here (`nvidia-cudnn-frontend` has no macOS wheel). `uv sync --no-default-groups` succeeds, so the suite *can* be exercised. Doing that surfaced **two genuine regressions this PR was introducing**, both now fixed.

### Regression 1 — `cohere` 5.18.0 → 7.0.8 broke the reranker tests

cohere 7 resolves submodules lazily via a module-level `__getattr__`, and `client_v2` is not in its `_dynamic_imports` table:

```
tests/autorag/nodes/passagereranker/test_cohere_reranker.py:50: in <module>
    @patch.object(cohere.client_v2.AsyncClientV2, "rerank", mock_cohere_reranker)
E   AttributeError: No client_v2 found in _dynamic_imports for module name -> cohere.
    Did you mean: 'ClientV2'?
```

Production was never affected — `autorag/nodes/passagereranker/cohere.py` already constructs `cohere.AsyncClientV2(...)`. Only the test reached through a private submodule path. Fixed by patching the public symbol; **3 passed under cohere 5.18.0 and under 7.0.8**, so the change is safe on both.

### Regression 2 — `llama-index` 0.14.8 → 0.14.23 silently dropped `llama-parse`

This one was a **production feature break**, not a test issue. `autorag/data/parse/llamaparse.py` does `from llama_parse import LlamaParse` and is registered in `autorag/support.py` as the `llamaparse` parser — but `llama-parse` was never declared in `pyproject.toml`. It only existed as a transitive dependency of `llama-index`, and the upgrade removed that edge.

```
# pre-upgrade lock
>>> from autorag.data.parse.llamaparse import llama_parse
llamaparse OK

# post-upgrade lock (before this fix)
>>> from autorag.data.parse.llamaparse import llama_parse
ModuleNotFoundError: No module named 'llama_parse'
```

Fixed by declaring `llama-parse>=0.6.21` directly. That is the right call regardless of this upgrade — a module imported by production code should never depend on being someone else's transitive dependency.

### Evidence after both fixes

| Check | Baseline (`main`) | This PR |
|---|---|---|
| `pytest --collect-only tests/autorag` | 364 collected, 29 errors | **364 collected, 29 errors** |
| Error module set | — | **byte-identical to baseline** |
| `test_util.py` + `test_strategy.py` + `test_langchain_chunk.py` | 41 passed | **41 passed** |
| `test_cohere_reranker.py` | 3 passed | **3 passed** |
| Security posture | — | unchanged: 61 fixed / 33 blocked upstream |

The 29 remaining collection errors are all missing optional extras (27 × `torch` from `[gpu]`, `PyPDF2` and `fitz` from `[parse]`) and are present identically on `main`.

Still worth running full CI on ubuntu before merge, since the `[gpu]` extra — which pulls `torch`, `vllm` and `transformers` — cannot be exercised on macOS at all.

PR #1402, opened 2026-08-08
perf(parser): skip encoding detection for pure ASCII sources (4.2x cold refresh)

HaD0Yun:
`chardet.detect()` dominates document reading, and it runs even when the answer is already determined.

Per 4KB file:

| step | cost |
|---|---|
| `iconv.decode(buffer, "utf8")` | 10 us |
| **`chardet.detect(buffer)`** | **672 us** |
| `.normalize("NFC")` | 1 us |

So ~98% of the per-file budget is spent deciding what encoding a file is in — including for plain ASCII files, where every plausible answer decodes to the same string.

## Change

Skip detection when the buffer is **pure ASCII**: at least 4 bytes, every byte in `0x01`–`0x7F`, no ESC.

## Why this cannot change output

For a buffer meeting that predicate, `chardet` can only return `ASCII`, `UTF-8`, `ISO-8859-x`, `Shift_JIS`, `Big5`, `EUC-JP`, `EUC-KR` or `GB18030`. The others are structurally excluded:

- UTF-16 / UTF-32 need a BOM or NUL bytes — NUL is excluded by the predicate
- ISO-2022 needs ESC — excluded
- `windows-125x` recognizers need C1 bytes (`0x80`–`0x9F`) — excluded

Every remaining candidate maps `0x01`–`0x7F` to the same code points, and NFC is the identity on that range. So the decoded string is identical by construction, not by luck.

Buffers shorter than 4 bytes deliberately stay on the legacy path: `chardet`'s UTF-32LE heuristic can claim UTF-32LE for short control-character buffers, and `iconv-lite` then drops the incomplete trailing unit, so the existing output for those is `""`. Preserving that keeps this a pure optimization rather than a silent behavior change.

## Effect

Cold refresh of a 200-document corpus, median of 5 runs, same corpus and machine, comparing this branch against a separate checkout of `main`:

| corpus | `main` | this branch | |
|---|---|---|---|
| ASCII | 175 ms | **42 ms** | **4.2x** |
| **Korean** | 199 ms | 200 ms | **1.0x** |

Korean documents are not pure ASCII, so they take the legacy path and see no change. Worth stating plainly given this project's primary corpus: this speeds up English/code/ASCII-markdown ingestion and leaves Korean ingestion where it was.

Widening the predicate to "any valid UTF-8" would help Korean, but it is **not** safe as a pure optimization: the current code follows `chardet` even when UTF-8 decodes cleanly, so short or ambiguous inputs could decode differently. That would be a semantics change needing a version bump and re-parse, so it is out of scope here.

## Verification

`test/parser/decode-text.test.ts` keeps a copy of the previous implementation and asserts both produce byte-identical output across boundary lengths (0–5), boundary byte values (`0x00`, `0x01`, `0x7F`, `0x80`), UTF-8 BOM, NUL bytes, CP949-encoded Korean, combining characters, invalid sequences, and seeded pseudo-random buffers.

Adversarial sweeps beyond the committed test found no divergence in 321,053 generated cases.

`test/bench/measure-decode.ts` reproduces the numbers above; it is not collected by vitest.

## Scope

This is one of three optimizations I measured; it is the only one that is fully independent, so it is offered on its own. The other two (skipping BM25 rebuilds for an unchanged corpus, and caching query-time state) share a fingerprint mechanism and a cross-process lock, and they change CLI exit-code semantics, so they are a separate and much larger conversation. Happy to open that separately if this direction is welcome.

### DIFF
--- a/src/parser/text.ts
+++ b/src/parser/text.ts
@@ -1,10 +1,24 @@
 import { detect } from "chardet";
 import iconv from "iconv-lite";
 
+// Text decoding and normalization applied to mirror content before BM25 chunking: the stored
+// artifact contains exactly the text produced here, so a change invalidates every stored index.
 const WINDOWS_949_LABELS = new Set(["EUC-KR", "ISO-2022-KR", "windows-949", "CP949"]);
 
 export function decodeText(bytes: Uint8Array): string {
 	const buffer = Buffer.from(bytes);
+	if (isPureAscii(bytes)) {
+		// Bytes 0x01-0x7F decode to the same code points under utf8, cp949, and every single-byte
+		// encoding chardet can name, and NFC normalization is the identity on them, so the result
+		// is fixed without consulting chardet. ESC (0x1B) is excluded: chardet can claim
+		// ISO-2022-JP/CN for buffers containing its escape sequences, which iconv-lite cannot
+		// decode, so those bytes must keep the legacy path. Buffers shorter than one 4-byte unit
+		// are excluded too: chardet's UTF-32LE heuristic can claim UTF-32LE for control-character
+		// and DEL bytes, and iconv-lite then drops the incomplete trailing unit, making the legacy
+		// output "". Pure printable ASCII decodes normally ("A"/"AB"/"ABC"), so excluding every
+		// sub-4-byte buffer is just the conservative choice and must keep the legacy path.
+		return iconv.decode(buffer, "ascii");
+	}
 	const utf8 = iconv.decode(buffer, "utf8");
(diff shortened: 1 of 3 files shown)


vkehfdl1 (approved):
검토 완료했습니다. ASCII fast path는 기존 출력과 동등하며, 한국어 UTF-8/CP949 및 Shift_JIS, Big5, GB18030, Windows code pages, UTF-16 BOM 입력을 main 동작과 직접 대조했습니다. 로컬 전체 테스트 1,263개, typecheck, build, Biome, CI 모두 통과했습니다.

PR #1410, opened 2026-08-13
fix(legacy): preserve chat prompt roles

### LINKED ISSUE #1234: [BUG] Chat prompt truncation and gpt-5 routing drop role/turn structure
**Describe the bug**
Three spots in the generator layer collapse a chat prompt (`List[Dict]` of role/content messages) into a single string before it reaches the model, losing system/user/assistant structure and sometimes injecting raw chat-template control tokens into the content:

1. `autorag/nodes/generator/openai_llm.py::truncate_by_token` (lines 337-354 on `main`) — When a list is passed, it delegates to `tiktoken_messages_to_string`, which wraps every message with `<|im_start|>{role}\n{content}<|im_end|>`, joins with newlines, and returns a single `str`. The caller (`OpenAILLM._pure` and `astream`) then passes that string into `parse_prompt`, which rewraps it as a single `user` message containing the raw chat-template tokens.
2. `autorag/nodes/generator/vllm_api.py::VllmAPI.truncate_by_token` (lines 87-95) — For a list input, the method newline-joins only `msg["content"]` values, discarding all role information.
3. `autorag/nodes/generator/openai_llm.py::get_result_gpt_5` (lines 309-324) — After `parse_prompt(prompt)`, only `role == "system"` and `role == "user"` messages are kept; each group is `"\n\n".join(...)`-ed into two strings, so prior assistant turns and any `tool` messages are silently dropped.

For contrast, `autorag/nodes/generator/minimax_llm.py::truncate_messages_by_token` (lines 185-230) already implements the intended behaviour — it preserves the `List[Dict]` structure and truncates only inside the last message. Lifting that helper into a shared utility and using it from the three sites above would be the smallest change that fixes all three.

**To Reproduce**
Any chat prompt that exceeds `max_token_size`, or any gpt-5 call with a multi-turn prompt that includes assistant history. Minimum reproducible example:

```python
from autorag.nodes.generator.openai_llm import tiktoken_messages_to_string

msgs = [
    {"role": "system", "content": "You are helpful."},
    {"role": "user", "content": "Hello"},
    {"role": "assistant", "content": "Hi! How can I help?"},
    {"role": "user", "content": "What's the weather?"},
]
s = tiktoken_messages_to_string(msgs)
print(s)
# <|im_start|>system
# You are helpful.<|im_end|>
# <|im_start|>user
# Hello<|im_end|>
# <|im_start|>assistant
# Hi! How can I help?<|im_end|>
# <|im_start|>user
# What's the weather?<|im_end|>
# <|im_start|>assistant
```
This `s` is then handed back to `parse_prompt`, which re-wraps it as `[{"role": "user", "content": s}]` — the four-turn conversation is gone and control tokens are now part of the user content.

**Expected behavior**
Truncation preserves the `List[Dict]` structure; gpt-5 dispatch sends the full multi-turn messages to `client.responses.create` as structured input instead of flattening roles into two strings.

**Full Error log**
No exception is raised — this is a silent correctness bug. The model receives garbled control tokens (openai/vllm case) or a message history missing its assistant turns (gpt-5 case), which degrades answers.

**Code that bug is happened**
See files and line ranges above. The MiniMax helper `truncate_messages_by_token` is the target pattern.

 - OS: macOS 15.3
 - Python version: 3.12

**Additional context**
PR #1194 ("gpt-5 support") touches `get_result_gpt_5`, so it would be good to agree on scope before sending a refactor: (a) generalise `truncate_messages_by_token` in `utils/util.py`, (b) switch `openai_llm.py` and `vllm_api.py` to use it, (c) feed structured messages through `get_result_gpt_5` instead of collapsing them. Happy to send a focused PR if the plan is acceptable.

yaodong-shen:
## Summary

- preserve structured chat messages during token-budget truncation for the legacy OpenAI, MiniMax, and VLLM API generators
- send the complete role-preserving message list to the GPT-5 Responses API instead of flattening system/user turns and dropping assistant history
- retain the existing MiniMax helper import surface while sharing the truncation implementation

## Root cause

The OpenAI and VLLM API paths converted `List[Dict]` chat prompts to a single string before dispatch. That turned role markers into user content; GPT-5 additionally discarded prior assistant turns when forming a Responses request.

## Validation

- `PYTHONPATH=/private/tmp/autorag-test-stubs uv run --no-sync pytest tests/autorag/nodes/generator/test_openai.py tests/autorag/nodes/generator/test_minimax.py tests/autorag/nodes/generator/test_vllm_api.py -q -k 'truncate_chat_prompt_preserves_message_roles or gpt_5_responses_preserve_full_chat_history or truncate_messages_preserves_roles or vllm_api_truncate_preserves_chat_messages'`
  - 4 passed
- `uvx ruff check` on all changed generator and test files
- `uvx ruff format --check` on all changed generator and test files
- `git diff --check`

The full legacy suite was not runnable locally because its macOS ARM development resolution includes an NVIDIA-only package and its OpenAI test module imports optional `torch.cuda` during collection. The focused regression tests use no model call, credentials, or GPU.

Closes #1234

### DIFF
--- a/legacy/autorag/nodes/generator/chat.py
+++ b/legacy/autorag/nodes/generator/chat.py
@@ -1,0 +1,72 @@
+"""Helpers for truncating chat prompts without changing their message roles."""
+
+from typing import Callable, Dict, List, Sequence, TypeVar, Union
+
+
+Token = TypeVar("Token")
+Encode = Callable[[str], Sequence[Token]]
+Decode = Callable[[Sequence[Token]], str]
+Prompt = Union[str, List[Dict]]
+
+
+def messages_to_string(messages: List[Dict[str, str]]) -> str:
+	"""Render messages only for token counting, never for model input."""
+	formatted_parts = [
+		f"<|im_start|>{message['role']}\n{message['content']}<|im_end|>"
+		for message in messages
+	]
+	formatted_parts.append("<|im_start|>assistant")
+	return "\n".join(formatted_parts)
+
+
+def count_message_tokens(messages: List[Dict], encode: Encode) -> int:
+	"""Count the tokenized representation of a chat prompt."""
+	return len(encode(messages_to_string(messages)))
+
+
+def truncate_messages_by_token(
+	messages: List[Dict], encode: Encode, decode: Decode, max_token_size: int
+) -> List[Dict]:
+	"""Fit a chat prompt in a token budget while preserving message structure."""
+	truncated_messages = [dict(message) for message in messages]
+	if count_message_tokens(truncated_messages, encode) <= max_token_size:
+		return truncated_messages
+
+	result: List[Dict] = []
+	for message in truncated_messages:
+		candidate = result + [message]
+		if count_message_tokens(candidate, encode) <= max_token_size:
+			result.append(message)
+			continue
+
+		content_tokens = encode(message.get("content", ""))
+		low, high = 0, len(content_tokens)
+		best_content = ""
+		while low <= high:
+			mid = (low + high) // 2
+			truncated_message = {
+				**message,
+				"content": decode(content_tokens[:mid]),
+			}
+			candidate = result + [truncated_message]
+			if count_message_tokens(candidate, encode) <= max_token_size:
+				best_content = truncated_message["content"]
+				low = mid + 1
+			else:
+				high = mid - 1
+
+		if best_content or not result:
+			result.append({**message, "content": best_content})
+		break
+
+	return result
+
+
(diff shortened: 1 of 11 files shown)


vkehfdl1 (approved):
Reviewed the full legacy generator diff. Structured chat roles and assistant history are preserved across OpenAI, GPT-5 Responses, MiniMax, and VLLM API paths; existing Vllm and LlamaIndex chat paths remain compatible. Verified with locked UV sync, 62 passing generator tests, targeted ty checks, Ruff lint, and format checks.

PR #1423, opened 2026-08-17
fix(minsync): stage parsed mirrors incrementally

vkehfdl1:
## Summary
- preserve the MinSync staging tree and copy only added or changed parsed mirrors
- sweep deleted and crash-orphaned staged files while rejecting unsafe state keys
- add RED→GREEN coverage for modified, add/delete, traversal, and interrupted-staging cases

## Verification
- `bunx vitest run test/minsync/minsync.test.ts` — 22 passed
- `bunx vitest run --exclude test/subagents/runtime.test.ts` — 89 files, 1266 tests passed
- `bun run typecheck`
- `bun run build`
- `bunx biome check src/minsync/workspace.ts test/minsync/minsync.test.ts`
- live E2E with real MinSync v0.3.0: initial sync processed 2 files; one-file update processed 1; unrelated inode/mtime/hash unchanged; only changed policy reached TEI

## Validation note
The isolated worktree exposes an unrelated `test/subagents/runtime.test.ts` assumption that constructs `process.cwd()/node_modules`; its focused test passes on the normal main checkout. A concurrent file-lock test also passed when rerun alone after one full-suite flake.

### DIFF
--- a/src/minsync/client.ts
+++ b/src/minsync/client.ts
@@ -125,7 +125,7 @@ function readCheckFailure(stdout: string): string | undefined {
 
 function parseQueryHits(stdout: string): readonly MinSyncQueryHit[] {
 	const parsed = parseJson(stdout);
-	const candidates = isRecord(parsed) ? parsed.results : parsed;
+	const candidates = Array.isArray(parsed) ? parsed : isRecord(parsed) ? parsed.results : [];
 	if (!Array.isArray(candidates)) return [];
 	return candidates.filter(isMinSyncQueryHit);
 }
--- a/src/minsync/workspace.ts
+++ b/src/minsync/workspace.ts
@@ -2,6 +2,7 @@ import { randomUUID } from "node:crypto";
 import {
 	copyFileSync,
 	existsSync,
+	lstatSync,
 	mkdirSync,
 	readdirSync,
 	readFileSync,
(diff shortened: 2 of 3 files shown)


vkehfdl1:
Blocking filesystem-integrity regressions found during review:

1. `src/minsync/workspace.ts:58-85,124-131`: after a valid state exists, a symlinked `workspace/files` root is preserved. `removeUnexpectedStagedFiles()` follows that directory symlink, writes staged mirrors into its target, and deletes target files not in `desiredPaths`. Local reproduction: `victimExists:false`, `outsideWrite:true`. Refuse/remove a symlinked files root before copying or cleanup, and add a regression test proving the link target survives.

2. `src/minsync/workspace.ts:68-81`: a staged leaf symlink is treated as an existing unchanged file; MinSync can read the external target. When `updatedAt` changes, `copyFileSync(..., entry.minSyncPath)` follows the leaf symlink and overwrites the external target. Stage to a temporary regular file and atomically rename over the destination, while rejecting/recreating symlinked parents/targets; add a regression test.

Focused MinSync tests (22), typecheck, build, Biome, CI and CodeQL otherwise pass, but these data-loss/arbitrary-write cases block merge.

vkehfdl1:
Fixed the blocking symlink issues in `dfd7439f`:

- managed staging roots and parent directories are checked with link-aware filesystem metadata and rebuilt without traversing link targets
- staged files are copied to a temporary regular file and atomically renamed over the destination
- unchanged entries are preserved only when the staged destination is a real regular file
- real MinSync v0.3.0 top-level array query responses are now accepted

Verification: 24/24 focused tests; 89 files / 1268 tests excluding the known isolated-worktree runtime assumption; typecheck, Biome and build passed. Live E2E with real MinSync v0.3.0 + local TEI: initial sync 2 files, one-file update 1, unchanged inode/mtime preserved, both symlink targets survived, and the updated policy was returned by semantic query.

PR #1480, opened 2026-08-26
fix: abort searches that exceed the configured timeout

### LINKED ISSUE #1475: Bound AutoRAG agent searches when model/tool loop never emits a terminal event
## Summary
`searchDocuments()` can remain pending indefinitely when the Pi agent/model stream or tool loop never emits a terminal event. In the live GLM-5.2 Fireworks validation, two parallel searches stayed at `search_started` for more than three minutes with 0% CPU and no `search_completed` or `search_failed` event. They had to be killed externally.

## Reproduction evidence
- Model: `accounts/fireworks/routers/glm-5p2-fast`
- Provider: Fireworks OpenAI-compatible endpoint
- Query A: `AutoRAG Fireworks GLM-5.2 model configuration`
- Query B: `Discord 회의 일정 프로젝트`
- Observed: process remained alive >3 minutes, no terminal run event
- Control query `지원사업 중간점검 제출서류` completed in about 125 seconds

## Suspected cause
The AutoRAG agent construction/run path does not set an overall abort deadline, a maximum tool/agent iteration count, or a guaranteed terminal-event fallback around the Pi agent loop. A stalled model stream or tool call therefore leaves the promise unresolved.

## Expected
- Add a configurable, moderately generous overall timeout (for example 10 minutes).
- Add a configurable max iteration/tool-turn limit, with a higher default appropriate for multi-source retrieval.
- Always emit `search_failed` with a timeout/iteration diagnostic and settle the promise on abort.
- Add deterministic tests for stalled stream, stalled tool, timeout, and iteration-limit paths.
- Keep partial trace/timing information for bottleneck analysis.

## Environment
- macOS arm64
- AutoRAG main at `8b44f5e5`
- GLM-5.2 via Fireworks

## Related observation
The same search path can spend around two minutes in a successful multi-step agent loop, so the timeout should be generous rather than a short request timeout.

mikemikimike:
## Summary
- add a configurable overall timeout for `searchDocuments()` (default: 10 minutes)
- abort the active Pi session when the timeout fires
- cap retrieval/tool executions per search (default: 32)
- validate both safety bounds and preserve the existing failure/cleanup path

## Issue
Closes #1475

## Validation
- `bun test test/observability/run-log.test.ts` — passed
- `bun test test/agent/agent.test.ts` — blocked locally because dependencies such as `typebox` are not installed
- `bun test` — repository baseline CI passed remotely for build and macOS/Windows compatibility; local run had missing dependencies and pre-existing Bun/Vitest timer API failures
- `bun run typecheck` / Biome — blocked locally because dependencies/tools are unavailable
- `git diff --check` — passed

`bun install --frozen-lockfile` was attempted but rejected because the repository lockfile is not synchronized with the package manifest; no lockfile changes were made.

### DIFF
--- a/src/agent/agent.ts
+++ b/src/agent/agent.ts
@@ -202,6 +202,8 @@ export interface AutoRAGAgentOptions {
 	excludeExactDuplicates?: boolean;
 	datasourceSkills?: readonly DatasourceSkill[];
 	datasourceAccess?: DatasourceAccessContextOptions;
+	/** Maximum time a model/tool search may run before it is aborted. */
+	searchTimeoutMs?: number;
 }
 
 export interface AutoRAGSearchSession {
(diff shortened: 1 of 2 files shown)


vkehfdl1 (approved):
Reviewed. The default retrieval/tool call limit is now 128, with regression coverage; CI is green.

PR #1497, opened 2026-08-30
fix(minsync): expose configurable chunk size

### LINKED ISSUE #1495: bug(minsync): expose chunk size and preserve it for local embedding context limits
## Summary

AutoRAG cannot configure MinSync chunk size through its public config, and each refresh initializes a MinSync workspace with the default `max_chunk_size = 4096`. This breaks local embedding models with smaller context windows such as Ollama `embeddinggemma:latest` (2048-token context).

## Reproduction

1. Configure AutoRAG with MinSync embedder `tei:embeddinggemma:latest`, base URL `http://127.0.0.1:18080`, dimension 768.
2. Run `autorag refresh --method minsync --json` over the configured Downloads/Documents/Desktop folders.
3. The generated `.minsync/config.toml` contains `max_chunk_size = 4096`.
4. The TEI adapter forwards chunks to Ollama; Ollama returns HTTP 500: `the input length exceeds the context length`.
5. Manually change `max_chunk_size` to 1000 and sync succeeds for the same corpus.

## Observed

The embedder settings are configurable, but chunking is not. A manual `.minsync/config.toml` edit is overwritten by a later AutoRAG refresh/init.

## Expected

Expose a validated MinSync chunker option in AutoRAG config (or derive a safe chunk budget from the embedder), preserve it during refresh, and surface an actionable diagnostic when the configured chunk budget exceeds the embedder context.

## Evidence

- EmbeddingGemma smoke request at 1000 chars: 768-dimensional vector.
- Request at 4000 chars: HTTP 500 `the input length exceeds the context length`.
- AutoRAG generated config default: `max_chunk_size = 4096`.
- No duplicate AutoRAG issue found.

mikemikimike:
## Summary

Fixes #1495 by making MinSync chunk-size configuration safe to use with the current defaults.

- Preserve MinSync's verified auto-install default when `autorag init --minsync-max-chunk-size` is used.
- Return the same exit code (`2`) used by other configuration errors for invalid chunk sizes.
- Restore the previous MinSync TOML when a forced rebuild cannot complete, avoiding a new chunk-size setting paired with an old cursor.
- Document `minSync.maxChunkSize` and retain the verified-release auto-install guidance.

## Implementation

`MinSyncClient` snapshots the configuration before applying embedder or chunk-size overrides and restores it on preflight, sync, or cursor-validation failure. A changed chunk size continues to force a full sync; successful rebuilds keep the new setting. The CLI now merges the chunk-size flag without overriding `autoInstall`, and validates the flag through the normal `ConfigError` path.

## Compatibility

Existing configurations and explicit `autoInstall: false` settings are preserved. No new runtime or configuration dependency is introduced.

## Validation

- `bun install --frozen-lockfile` — passed.
- `bunx vitest run test/minsync/minsync.test.ts -t 'forces a full sync when the configured chunk size changes|restores the previous config when a forced full sync fails|uses the MinSync chunk size for BM25-only indexing'` — 3 passed.
- `bunx vitest run test/cli/commands.init.test.ts -t 'MinSync chunk size|invalid MinSync chunk size'` — 2 passed.
- `bun run lint` — passed; Biome reported only the existing schema-version informational notice.
- `bun run typecheck` — passed.
- `bun run build` — passed.
- `git diff --check` — passed.
- GitHub Actions `build`, `compatibility (macos-latest)`, and `compatibility (windows-latest)` — passed.

The full `bun run test` suite was also started locally with the published MinSync Windows binary available. It reached `test/agent/bash-tool.test.ts`, where six tests failed because this Windows host's `bash.exe` forwards to WSL without `/bin/bash`; the suite stopped there. The MinSync test file also has two local Windows symlink setup failures (`EPERM`) before the test bodies run. The platform CI jobs above pass.


vkehfdl1 (dismissed):
Thanks for exposing the MinSync chunk-size setting. The configuration boundary is useful and the focused tests are a good start, but I found two functional gaps that should be addressed before merging.

1. **Changing `maxChunkSize` does not reindex an existing corpus (P1).**

`src/minsync/embedder-config.ts:75` updates `[chunker.options].max_chunk_size`, but `src/minsync/client.ts:94` still chooses incremental `sync` whenever `.minsync/cursor.json` exists. MinSync's cursor does not appear to include chunker option values, so unchanged files can retain their old chunks/embeddings after a user changes (for example) `4096` to `1000`. That means the context-overflow problem can persist even though the TOML shows the new value. Please either detect a chunk-size change and force a full reindex, or make the configuration change fail/actionably require a rebuild. Please also add a regression test for an already-indexed workspace.

2. **`--method bm25` bypasses the configured chunk size (P1).**

`maxChunkSize` is carried under `options.minSync` in `src/cli/config.ts:710-723`, while BM25 is constructed from the separate `options.bm25` object in `src/agent/agent.ts:401-415`. The BM25-only refresh path calls the BM25 method directly (`src/agent/agent.ts:1022-1024`), so a fresh workspace initialized with `autorag refresh --method bm25` can use MinSync's default chunk size instead of `minSync.maxChunkSize`. Please ensure BM25, vector, and hybrid paths share the same MinSync chunk-size configuration, and add a regression test covering a BM25-only refresh.

Once these two paths are covered, I would be happy to re-review.

vkehfdl1 (dismissed):
The two previously requested functional gaps look addressed on HEAD 253c4ab8: a chunk-size change with an existing cursor forces `minsync sync --full`, and BM25 now inherits `minSync.maxChunkSize`. Focused tests, typecheck, and build pass.

Do **not** merge yet. This branch still reverts #1505's auto-install default.

## Blocking

1. **`autorag init --minsync-max-chunk-size N` writes `autoInstall: false`.**
   `src/cli/commands/init.ts` sets `partial.minSync = { enabled: true, autoInstall: false, maxChunkSize }`. Embedder flags later overwrite that to `true`, so:
   - chunk-size only → `autoInstall: false` (users lose auto-install)
   - chunk-size + embedder flags → `autoInstall: true`
   Omit `autoInstall` from this partial (or set `true`) so #1505's default is preserved. Add an init test for the chunk-size-only path.

2. **README.md reverts the #1505 auto-install sentence.**
   Main currently documents auto-install of a verified release (`autoInstall: true`). This PR replaces that with "MinSync uses a pre-installed binary (`autoInstall: false`)" while adding `maxChunkSize`. Keep the #1505 wording and add the chunk-size note to it. This will also conflict when updating the branch (behind main by #1505).

## Non-blocking

- Invalid `--minsync-max-chunk-size` throws and exits 1; other init config errors return 2. Handle `ConfigError` like the embedder-flag path.
- `rewriteEmbedderConfig` writes the new chunk size before `sync --full`. If that full sync fails, the next run sees a matching TOML value plus the old cursor and falls back to incremental sync. Persist-or-rebuild should not commit the new size until the rebuild succeeds (or delete the cursor when forcing rebuild).
- `docs/minsync-setup.md` still does not mention `minSync.maxChunkSize`.

mikemikimike:
Addressed the review feedback in commit 4cb8e70f:

- Chunk-size-only init now preserves the verified MinSync auto-install default (`autoInstall: true`), and invalid values use the standard configuration-error exit code (`2`).
- MinSync snapshots the previous config and restores it if preflight, the forced rebuild, or cursor validation fails, preventing a new chunk size from being paired with an old cursor.
- README and `docs/minsync-setup.md` now retain the verified-release guidance and document `minSync.maxChunkSize`.

Targeted MinSync/CLI tests, build, lint, and typecheck pass locally. The full local suite remains limited by this Windows host's missing WSL `/bin/bash` and symlink permissions; all three GitHub CI jobs (build, macOS compatibility, and Windows compatibility) pass.

vkehfdl1:
Addressed the remaining review findings on :\n\n-  no longer writes ; the #1505 default stays true. Invalid values now return exit 2.\n- README keeps the auto-install-by-default wording and adds .\n-  documents  / , including the EmbeddingGemma example.\n- Changing chunk size deletes  before rebuild, so a failed  cannot fall back to incremental.\n- Added a BM25-only  regression test.\n\nBranch is updated from .

vkehfdl1:
Follow-up on the actual PR head (`mikemikimike:pr-1495`), not `Marker-Inc-Korea/pr-1495`.

`4cb8e70f` already preserved `#1505` auto-install, returned exit 2 for invalid `--minsync-max-chunk-size`, restored README auto-install wording, and rolled back MinSync config after a failed full rebuild.

Pushed `1a45daee` on top:
- expanded `docs/minsync-setup.md` with `maxChunkSize` examples, including the EmbeddingGemma init command
- added a BM25-only `refresh({ methods: ["bm25"] })` regression test
- merged `main` (`#1509`) so the branch is current

vkehfdl1 (approved):
Reviewed the completed implementation and CI fix. All required checks pass; approving for merge.

PR #1513, opened 2026-09-02
fix(minsync): invalidate cursor before chunk rebuild

mikemikimike:
## Summary

Follow-up to #1497. When minSync.maxChunkSize changes, invalidate the existing MinSync cursor before the forced full rebuild. If the rebuild fails or does not produce a new cursor, keep the cursor invalidated so a later run cannot incorrectly use incremental sync with the old index state.

## Validation

- unx vitest run test/minsync/minsync.test.ts -t 'forces a full sync when the configured chunk size changes|restores the previous config when a forced full sync fails|uses the MinSync chunk size for BM25-only indexing' — passed.
- unx vitest run test/retrieval/bm25.test.ts — 11 passed.
- un run build — passed.
- un run typecheck — passed.
- Changed-file Biome check and git diff --check — passed.
- The broader local suite remains limited by unrelated 60-second agent lifecycle/tool-call-limit test timeouts on this Windows host; GitHub CI for #1497 passed on build, macOS, and Windows.

### DIFF
--- a/src/minsync/client.ts
+++ b/src/minsync/client.ts
@@ -1,4 +1,4 @@
-import { existsSync, readFileSync, writeFileSync } from "node:fs";
+import { existsSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { configuredMaxChunkSize, minSyncConfigPath, rewriteEmbedderConfig } from "./embedder-config.ts";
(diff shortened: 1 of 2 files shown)


vkehfdl1 (approved):
Reviewed the MinSync cursor invalidation change. Focused tests, typecheck, build, lint, and all required CI checks pass.

Issue #1560, opened 2026-09-13
Stale-index diagnostics do not say which source is stale or why


vkehfdl1:
**Problem.** When retrieval runs on a stale index, the only diagnostic is:

```
{ "code": "index-not-ready", "message": "Index is stale. Run `autorag lite refresh` or `autorag refresh` before retrieving." }
```

It does not say *which* source changed, *why* it is considered stale, or any action other than "run refresh". When staleness recurs (e.g. a tool writes a file into a `searchPaths` folder on every refresh), users repeat `refresh` in a loop without ever learning the cause.

**Repro (from a fresh sandbox).**
1. `autorag init --search-paths /tmp/demo/docs --workspace /tmp/demo/ws`
2. `autorag refresh` → ok
3. `autorag status` → `stale: true`, no diagnostic details
4. `autorag lite retrieve "query"` → `index-not-ready`, exit non-zero

**Expected.** The stale diagnostic should include: the affected source paths, the reason (mtime/size changed vs index), and actionable guidance (e.g. which file keeps being rewritten), so a user who is not familiar with index internals can fix the cause, not the symptom.

PR #1564, opened 2026-09-14
fix(defaults): local EmbeddingGemma embedder, cwd-safe init, dot-file scan skip, cross-process staleness

vkehfdl1:
## Summary

Batch of UX-defaults fixes discovered by running a fresh sandbox setup→index→search round as a non-RAG user would experience it.

- **fix(init): require an explicit absolute `--workspace`** — `writeDefaultConfig` no longer defaults `workspacePath` to cwd; `init` rejects relative `--workspace` with exit 2 and a clear message. Legacy config migration still supplies values.
- **fix(mirror): skip dot-prefixed files during source scans** — `.jikji_agent_map.md`-style index byproducts (files, not just `.jikji/` dirs) are excluded from parsed mirrors and the staleness scan.
- **fix(status): derive staleness from the stat scan only** — `stale` was pinned true by `lastOutcome === "never"` on every fresh CLI process (status runs in a new agent each time), producing an endless stale loop and hard-refusing `lite retrieve` right after a successful refresh. Now stat-based `detectMirrorStaleness` alone decides; a separate-agent-instance regression test is pinned.
- **feat(minsync): default to local EmbeddingGemma when no embedder is configured** — `MinSyncClient` injects `tei:embeddinggemma:latest` via the loopback TEI adapter (`127.0.0.1:18080`, dim 768) instead of falling through to upstream's `openai:text-embedding-3-small`. Corpus text stays local by default.
- **chore(minsync): pin minsync 0.4.4** — bumped to the latest crates.io release.

## Verification

- Fresh sandbox QA (no embedder flags, `OPENAI_API_KEY` unset): `init` → `refresh` → semantic `lite retrieve` hits the fixture document; `status` stays `stale: false` across invocations; `init` without/with-relative workspace exits 2 with the expected message.
- RED→GREEN for every change: new failing tests captured first (init validation, dot-skip, cross-instance stale, default embedder init args), then smallest production change.
- Full suite: all 115 test files passed.

Follow-up tracked in #1560 (stale diagnostics still do not name the stale source).

Test plan: covered above; environment note — `test/cli/commands.refresh-status.test.ts` is environment-sensitive when a live TEI adapter plus a real minsync on PATH are both present (sync genuinely succeeds → component reads `ready`); passes on a machine without the adapter running.


vkehfdl1:
Closed per maintainer direction: the MinSync default-embedder portion of this PR overlaps with and is superseded by https://github.com/Marker-Inc-Korea/AutoRAG/pull/1574 (shared embedding runtime: autorag-gateway + MinSync identity fingerprint). The orthogonal fixes here (init workspace guard, dot-file scan skip, cross-process staleness) are noted in the superseding PR's closing thread so they are not silently lost — they can be re-proposed independently if still wanted.

PR #1600, opened 2026-09-17
fix(embedding-runtime): accept MinSync default batches and keep batching settings

vkehfdl1:
## Why

A fresh MinSync workspace cannot sync through the AutoRAG-owned embedding gateway. Two compatibility bugs compound into one silent failure.

**1. The gateway rejects the largest request MinSync's own defaults can build.**
MinSync 0.4.x inits a fresh workspace with `batch_size = 64` (`[embedder]`) and `max_chunk_size = 4096` (`[chunker.options]`), so one embedding request can carry `64 * 4096 = 262,144` characters. The gateway capped a request at `100,000` characters, so every request was refused:

```
Error: embedding failed: TEI API error 413 Payload Too Large:
{"error":"Embedding batch exceeds the configured limit.","code":"batch-limit"}
```

Reproduced against a real workspace: the MinSync profile sync exits with code 5, the LanceDB store is left half-written with an open `txn.json`, and no `cursor.json` is ever created.

**2. A profile-selected runtime drops the operator's batching settings.**
`effectiveEmbedder()` resolved a profile to `{id, baseUrl, dimension, queryPrefix, passagePrefix}` only. `batchSize`, `maxRetries`, `maxConcurrent`, and `timeoutMs` from `minSync.embedder` never reached MinSync's `config.toml`, so those settings silently reverted to MinSync's defaults (batching 64) even when the operator had pinned something else.

Together these made a fresh install unable to build a semantic index, while `autorag refresh --json` still reported `{"ok": true}` (that part is a separate rendering issue and is not touched here).

## What changed

- `src/embedding-runtime/gateway.ts`: derive the default per-request character budget from MinSync's defaults (`64 * 4096`), with the derivation documented next to the constants; the existing 2 MB body cap already covers the worst-case JSON encoding of that budget.
- `src/minsync/client.ts`: carry `batchSize`, `maxRetries`, `maxConcurrent`, and `timeoutMs` into the runtime-resolved embedder config so the operator's values are written to MinSync's `config.toml`.
- Tests: a gateway test that the largest MinSync-default batch is accepted (and that the derived limits still reject oversized batches), plus a MinSync test that a profile config keeps operator batching settings in the rewritten config.

## Verification

- `vitest run test/embedding-runtime/gateway.test.ts test/minsync/minsync.test.ts` -> 56 passed
- `bun run typecheck` -> clean
- `biome check` on the four changed files -> clean
- Live check on a real workspace: with the pre-fix gateway a profile sync failed with the 413 above; after aligning the limit (and pinning `minSync.maxChunkSize` to the documented 1000 for a fresh workspace) a full rebuild embeds through the local gateway without a 413.


vkehfdl1:
Live evidence from the machine that hit this (AutoRAG 2.5.0, macOS arm64, local `qwen3-embedding-0.6b` gateway):

**Bug 1 (gateway cap).** A real full reindex over a corpus of ~1,600 parsed documents never reached MinSync's `check`/`sync`:

```
$ minsync check --format json        # after AutoRAG rewrote the workspace config for the gateway
Error: vector store error: dimension mismatch for LanceDB vector column: existing 768, requested 1024
```

and after resetting the legacy 768-dim store, the fresh store failed with the cap itself:

```
Error: embedding failed: TEI API error 413 Payload Too Large:
{"error":"Embedding batch exceeds the configured limit.","code":"batch-limit"}
```

The workspace was left half-written (`txn.json` present, no `cursor.json`), and `autorag refresh --json` still reported `{"ok": true}`.

**Bug 2 (dropped settings).** With `minSync.embedder.batchSize = 32` in `~/.autorag/config.json`, the workspace config AutoRAG writes for a profile-selected runtime still contained MinSync's own default:

```toml
[embedder]
id = "tei:Qwen3-Embedding-0.6B-Q8_0.gguf"
batch_size = 64      # configured 32 never arrives
```

With this branch the operator's `batchSize`/`maxRetries`/`maxConcurrent`/`timeoutMs` reach that file instead, and the derived gateway budget accepts the largest batch MinSync's defaults can build (`64 * 4096`).

vkehfdl1:
Definitive reproduction and fix verification on the real product path (macOS arm64, local `qwen3-embedding-0.6b` gateway, fixture = one 1.3 MB delimiter-rich document so MinSync builds its maximum default batch).

**Measured request shape** (fake TEI endpoint in front of `minsync sync --full` with MinSync's own defaults `batch_size = 64`, `max_chunk_size = 4096`):

```
requests: 6
max inputs: 64            # count cap (128) was never the constraint
max characters: 253,841   # over the pre-fix 100,000-character cap
requests over 100,000 chars: 5
```

So the binding limit is the per-request character budget, and the observed worst case (253,841) sits just under `64 * 4096 = 262,144` — the value this branch derives.

**A/B on `autorag` itself** (same fresh workspace, `maxChunkSize` unset so MinSync inits with 4096):

| run | command | cursor.json | minsync status |
|---|---|---|---|
| A (2.5.0, pre-fix) | `autorag refresh --method minsync` | MISSING (stale `txn.json` left behind) | `NotSynced`, `last_synced_at = null` |
| B (this branch) | `autorag index rebuild --method minsync --yes` | PRESENT | `UpToDate`, `last_synced_at = 2026-09-17T06:28:58Z`, `embedder = tei:Qwen3-Embedding-0.6B-Q8_0.gguf` |

Both runs exited 0 and A reported `{"ok": true}` — the failure is only visible through `minsync status` (tracked separately as the silent-refresh issue, not in this PR).

The same branch also re-synced the affected machine's real workspace (1,556 staged documents): `state: UpToDate`, `cursor.json` written, and `autorag search --json` now returns curated results grounded in the corpus.

PR #1601, opened 2026-09-17
fix(datasource): reap the katok process tree so the client cannot hang

askalf:
## Summary

- `spawnKatok` never passed `detached`, so the child was not a POSIX process-group leader and the group kill at `client.ts:258` (`process.kill(-child.pid, "SIGKILL")`) **always threw `ESRCH`**. The tree-kill intent was dead code on every path: timeout, abort, stdout cap, stderr cap.
- Consequence: a katok descendant that inherited the client's stdout/stderr pipes was never signalled, and Node emits `close` only once every holder of those pipes exits — so the result promise **stayed pending forever**, with no bounded fallback.
- Fix: spawn as a process-group leader on POSIX (`detached: process.platform !== "win32"`, `windowsHide: true`), route all four termination sites through `terminateProcessTree` from `src/process/terminate-tree.ts`, and add the `exit`-path tree reap plus the `TREE_KILL_GRACE_MS` settle guard — the same shape maintainers merged for qmd in #1582.
- The single `settleWith`/`finish` path also retires the `child.killed` termination gate, which was set when a signal was *sent* rather than when the process exited, making a second termination (abort after a stdout-cap kill) a no-op.
- **11 tests** in a new `process tree termination` block: 9 discriminating (each hangs for 10s and fails on base), 2 controls. No production behaviour outside the termination/settle path changes; the existing 15 katok client tests are untouched and still pass.

All counts and transcripts in this body are reconciled to head `abe6b7e395c717b0c275a30018c6d68ed0ac3ed8`.

```console
$ npx tsx repro.mjs        # on base 0602ff3
A: katok spawn options (no detached) + process.kill(-pid, "SIGKILL") -> threw ESRCH
A: detached: true                   + process.kill(-pid, "SIGKILL") -> succeeded
B: katok doctor never settled within 10s (waited 10007ms)
B: descendant 42679 -> SURVIVED (still holding the inherited pipes)

$ npx tsx repro.mjs        # with the fix
A: katok spawn options (no detached) + process.kill(-pid, "SIGKILL") -> threw ESRCH
A: detached: true                   + process.kill(-pid, "SIGKILL") -> succeeded
B: settled in 71ms -> ok=false reason=invalid-shape
B: descendant 42862 -> reaped

$ npx vitest run test/datasource/skills/katok/client.test.ts   # base source + all 11 new tests
 Test Files  1 failed (1)
      Tests  9 failed | 17 passed (26)

$ npx vitest run test/datasource/skills/katok/client.test.ts   # with the fix
 Test Files  1 passed (1)
      Tests  26 passed (26)
```

(In part B of the repro, `reason=invalid-shape` is the fake binary's payload not matching the doctor schema — the point is that it **settles in 71ms instead of hanging**, and the descendant is gone. The adversarial verification re-checked this: it is cosmetic to the claim, not a masked failure — the fixed arm settles in 71ms where base never settles at all.)

## Decisions

Minimal focused fix; alternatives and boundaries are documented in the change rationale and regression coverage.

Not run: full cross-platform/manual integration matrix; scoped regression and relevant static checks were run as recorded above.

AI assistance: this bug was found and the fix and tests were drafted with AI tooling in my workflow; the tests and checks above were executed as pasted. I'm responsible for the change and will handle review feedback.


vkehfdl1:
Closing without merge: the katok skill was retargeted to lazykatok in #1668, so this no longer applies to a file on main. The same fix — POSIX process-group spawn, terminateProcessTree on timeout/abort/cap, bounded TREE_KILL_GRACE_MS exit grace, Windows taskkill fallback — now lives in #1671 against the lazykatok client (ported via the #1624 variant of this change). #1671 carries one descendant-cleanup regression test rather than this PR's 11-test termination matrix.

PR #1623, opened 2026-09-18
fix(setup): stop requiring env credentials for CLI-backed datasources; drop the discrawl Discord bot-token path

### LINKED ISSUE #1614: [BUG] `autorag setup` datasource probe disagrees with refresh reality (discord reported skipped without a token it does not need)
## Summary
`autorag setup`'s datasource probe disagrees with what a refresh actually does: it reports `discord: skipped (credential DISCORD_BOT_TOKEN is unavailable)` while `autorag refresh` indexes the same discord datasource successfully from discrawl's native store without that token.

## Repro
Config (`~/.autorag/config.json`) has discord with an instance id and a `tokenEnv`, but no bot `source`:
```json
"discord": { "instanceId": "geumcheon-artville", "connector": { "tokenEnv": "DISCORD_BOT_TOKEN",
  "guildId": "1490970691820847155", "channels": ["일반"], "maxDocuments": 300 } }
```
```bash
autorag setup --json      # disk: {"name":"discord","state":"skipped","reason":"credential DISCORD_BOT_TOKEN is unavailable"}
autorag refresh --json    # {"ok": true, "skill": "discord", "instanceId": "geumcheon-artville", ...}
```
The discord lane indexes fine (`~/.autorag/workspace/.autorag/datasources/discord/geumcheon-artville/chunks.json`, 292K) and reports its own `FTS only` diagnostic — no token involved.

## Root cause
`src/cli/setup.ts` collects required credentials from `DEFAULT_CREDENTIALS` *and* from `connector.tokenEnv`/`apiKeyEnv`/`credentialEnv` (`configuredCredentialNames`), then skips the datasource when any of those variables is unset. For CLI-backed datasources that query a native store (`discrawl.db`), the token is only needed for the optional bot sync source, so the probe is stricter than the runtime.

## Impact
`setup` tells the operator a working datasource is unusable and recommends configuring an env var that is not required, while the datasource is simultaneously being indexed and served. Users cannot trust `setup`'s datasource report.

## Expected
Probe the same condition the runtime needs (native store present + binary present), and treat `tokenEnv` as required only for the connector modes that actually use it (e.g. discord bot source).

Environment: AutoRAG `main` @ `964b3138` (post-v2.5.0), macOS arm64 (Darwin 25.6.0), Node 26.8.2, MinSync 0.4.4, profile `qwen3-embedding-0.6b` (1024d, local gateway).

vkehfdl1:
## What

Closes the report in #1614 by making `autorag setup` probe what a refresh actually needs, and removes the Discord **bot-token** code that no part of the discrawl-backed datasource uses.

## Root cause (#1614)

`src/cli/setup.ts` collected required credentials from `DEFAULT_CREDENTIALS` **and** from `connector.tokenEnv`/`apiKeyEnv`/`credentialEnv`, then skipped the datasource when any was unset. A discord entry that carries `tokenEnv: "DISCORD_BOT_TOKEN"` was therefore reported as

```
discord: skipped (credential DISCORD_BOT_TOKEN is unavailable)
```

while `autorag refresh` indexed the same datasource from discrawl's native store without any token. (The old `connector.source === "bot"` special case was dead code too — `"bot"` was never a valid source value.)

## Changes

**`autorag setup` probe** — CLI-backed datasources (anything in `BUILTIN_BINARIES`) are no longer gated on env credentials: the external CLI owns its archive, index, and credentials. The probe now mirrors the runtime (binary present + native store present, plus the embedding model for semantic sources). The stale `SLACK_TOKEN` / `TELEGRAM_BOT_TOKEN` / `WHATSAPP_TOKEN` gates for slacrawl/telecrawl/wacrawl are gone for the same reason; a non-CLI connector such as `github` keeps its `GITHUB_TOKEN` gate.

**Discrawl client** — AutoRAG reads discrawl's own local archive (`wiretap` import + search/embed over the native store), so the bot path is deleted: the `"discord"`/`"both"` source selector, `DEFAULT_DISCRAWL_SOURCE`, `DiscrawlSourceKind`, the `--source`/`--guild` sync arguments, and the `DISCORD_BOT_TOKEN` child-env allowlist. `sync()` always drives `discrawl wiretap`; no Discord credential is configured, required, or forwarded. A legacy `connector.source` value in an operator config is simply ignored.

**Retained** (per this repo's datasource rules): the `DISCORD_USER_TOKEN` rejection guard — automating a user account violates Discord's Community Guidelines.

**Docs/QA** — README's Discord section and the live QA script header no longer instruct anyone to create a bot application or export a token.

## Breaking change

`DEFAULT_DISCRAWL_SOURCE` and `DiscrawlSourceKind` are removed from the public export surface, and `DiscrawlOptions.source` is gone (bot-backed archives are the CLI's own concern).

## Verification

- RED before the production change: `setup.test.ts` (discord probed as `skipped — credential DISCORD_BOT_TOKEN is unavailable`) and `client.test.ts` (child env carried `DISCORD_BOT_TOKEN`; `sync()` ran `--source discord --guild g1`). GREEN after.
- Full suite: `All 147 test files passed`; `biome check .` clean; `tsc --noEmit` clean.
- Real surface: the real `discrawl` binary through `DiscrawlClient` with no token (`doctor: ready`, `status: 3148 messages / 167 channels / 6 guilds`), and real `autorag setup --json` on an isolated `AUTORAG_HOME` + temp config with `tokenEnv: "DISCORD_BOT_TOKEN"` → `discord: blocked (embedding model is unavailable)`, `github: skipped (credential GITHUB_TOKEN is unavailable)`, zero `DISCORD_BOT_TOKEN` occurrences in the report.

Fixes #1614


vkehfdl1:
Review + live verification on macOS (this clone, real `discrawl` 0.15.1, no Discord credential of any kind):

- **Code**: the probe gate now keys off `BUILTIN_BINARIES`, so `slack`/`telegram`/`whatsapp` losing their `DEFAULT_CREDENTIALS` entries is pure dead-code removal (all three are CLI-backed and would never reach the credential branch), while the non-CLI `github` connector keeps `GITHUB_TOKEN`. A legacy `connector.source` in an operator config flows through the factory spread and is simply ignored by the client — no crash, no bot path.
- **Live discrawl e2e**: `doctor db=true fts=true embeddings=true model=bge-m3`; `wiretap` synced 2937 messages into discrawl's native store; archive `3163 messages / 168 channels / 6 guilds`. All three retrieval methods returned real hits for Korean and English queries (`discord-hybrid`, `discord-fts`, `discord-semantic`), each with a source-native identity such as `/discord/live/chunks/1550345217079640115`.
- **Token isolation on the real surface**: with `DISCORD_BOT_TOKEN=fake-...` exported, `DiscrawlClient.search` still succeeds and the token is not forwarded to the child; the `DISCORD_USER_TOKEN` rejection guard is untouched.
- **Real `autorag setup --json`** (isolated `AUTORAG_HOME`, config carrying `tokenEnv: "DISCORD_BOT_TOKEN"`): `discord -> blocked (embedding model is unavailable)` instead of the #1614 `skipped (credential DISCORD_BOT_TOKEN is unavailable)`, `github -> skipped (credential GITHUB_TOKEN is unavailable)`, zero `DISCORD_BOT_TOKEN` occurrences in the report.
- **Checks**: `tsc --noEmit` clean, `biome check .` clean for the changed files (only pre-existing `src/p2p/*` warnings), `All 147 test files passed`, CI green.

Note (pre-existing, not from this PR): the live QA script's closing "re-running index should sync 0 new messages" expectation does not hold — `discrawl wiretap` re-imports the desktop cache (second pass reported 2928). Worth a follow-up to the script's wording or to the incremental assertion.

PR #1624, opened 2026-09-18
fix(katok): terminate descendant processes on timeout

### LINKED ISSUE #1584: [BUG] katok datasource: process-group kill is unreachable — spawn lacks detached, so timeout kills only the direct child
**Describe the bug**

The katok datasource client intends to kill the whole process tree on timeout/abort/output-cap, but its POSIX process-group kill can never succeed: `spawnKatok` never sets `detached`, so the child is not a process-group leader and `process.kill(-child.pid, ...)` always throws `ESRCH`. Every termination silently degrades to signalling the direct child only.

**Code that bug is happened**

```ts
// src/datasource/skills/katok/client.ts:146-150 — no `detached`
const child = spawn(portable.command, [...portable.args], {
	env,
	...(request.cwd === undefined ? {} : { cwd: request.cwd }),
	stdio: ["ignore", "pipe", "pipe"],
});

// src/datasource/skills/katok/client.ts:254-263 — group kill that cannot fire
function terminate(child: ChildProcess): void {
	if (child.killed) return;
	if (process.platform !== "win32" && child.pid !== undefined) {
		try {
			process.kill(-child.pid, "SIGKILL"); // ESRCH: the child is not a group leader
			return;
		} catch {}
	}
	child.kill("SIGKILL");
}
```

**Impact**

- The tree-kill intent is dead code. Any katok descendant that inherits the client's stdout/stderr pipes keeps the result promise pending forever — the exact class fixed for qmd in #1578 / #1582.
- `child.killed` is set when a signal was *sent*, not when the process exited, so an abort after a stdout-cap termination is also a no-op.

**To Reproduce**

1. `rg -n 'detached' src/datasource/skills/katok/client.ts` → no match, while `process.kill(-child.pid, ...)` exists at line 258.
2. Force the branch: construct `KatokClient` with a fake `binaryPath` that spawns a long-lived child, trigger the timeout, and log the caught `ESRCH` from the group kill.

**Expected behavior**

Termination reaps the whole process tree on every platform, or reports that it could not.

**Suggested fix**

- spawn with `detached: process.platform !== "win32"` (POSIX group leader) and keep the Windows parent/child chain that `taskkill /t /f` walks;
- reuse `src/process/terminate-tree.ts` (`terminateProcessTree`) added in #1582, plus the `exit`-path tree reap and bounded settle guard from `src/datasource/skills/obsidian/client.ts`;
- gate termination on `child.exitCode === null && child.signalCode === null` instead of `child.killed`;
- add a regression test equivalent to "reaps a qmd descendant that keeps the inherited stdio open after the direct child exits".

**Additional context**

#1578 / #1582 fixed the identical class for qmd. This is an AutoRAG-side defect rather than an upstream one: katok installs as a single Mach-O binary (`~/.cargo/bin/katok`), so there is no launcher layer or upstream CLI change to ask for.

 - OS: macOS 26.0 (darwin arm64)
 - Node 26.8.2 / Bun 1.3.14
 - katok: cargo-installed binary at ~/.cargo/bin/katok

lorenzozanee:
Fixes #1584.

Katok timeout, abort, and output-cap termination now starts a POSIX process group and uses the shared process-tree termination helper, while preserving Windows taskkill tree handling. Exit cleanup has bounded grace for descendants retaining stdio pipes.

Added a regression covering descendant cleanup and bounded timeout settlement.

Tests:
- `bunx vitest run test/datasource/skills/katok/client.test.ts`
- `bunx vitest run test/datasource/skills/katok`
- `bun run typecheck`
- `bun run build`
- `bunx biome check src/process/terminate-tree.ts src/datasource/skills/katok/client.ts test/datasource/skills/katok/client.test.ts`


vkehfdl1:
Superseded by #1671, which ports this fix onto the current `lazykatok` skill after the katok→lazykatok retarget in #1668.

vkehfdl1:
Closing without merge: the katok skill was retargeted to lazykatok in #1668, and this fix now lives in #1671.

PR #1625, opened 2026-09-18
fix(discrawl): bound process settlement after child exit

### LINKED ISSUE #1585: [BUG] discrawl datasource: no post-exit tree reap or bounded settle guard (#1578 class)
**Describe the bug**

The discrawl client terminates the process tree on timeout, abort, and output-cap, but it has no post-exit handling: the promise settles only on the child's `close` event. If the direct child exits while a descendant still holds the inherited stdout/stderr pipes, `close` never fires and the operation hangs — the #1578 class, which was fixed for qmd in #1582 but not for discrawl.

**Code that bug is happened**

```ts
// src/datasource/skills/discrawl/client.ts:240-245 — group leader is set up correctly
const child = spawn(portable.command, [...portable.args], {
	env,
	...(request.cwd === undefined ? {} : { cwd: request.cwd }),
	detached: process.platform !== "win32",
	stdio: ["ignore", "pipe", "pipe"],
});

// src/datasource/skills/discrawl/client.ts:369-378 — tree kill only on request
function terminate(child: ChildProcess): void {
	if (child.killed) return;
	if (process.platform !== "win32" && child.pid !== undefined) {
		try {
			process.kill(-child.pid, "SIGKILL");
			return;
		} catch {}
	}
	child.kill("SIGKILL");
}
```

There is no `child.on("exit")` handler, no bounded settle guard, and the result resolves only from the `close` listener.

**Gap versus the fixed qmd client (#1582)**

| | qmd (`obsidian/client.ts`) after #1582 | discrawl today |
|---|---|---|
| POSIX process group | yes | yes |
| tree kill on timeout/abort/cap | yes | yes |
| tree reap when the direct child exits | yes | **no** |
| bounded settle guard (no infinite pending) | yes | **no** |
| termination gate | `exitCode/signalCode` | `child.killed` |

**Impact**

- Any discrawl descendant holding the inherited pipes keeps the client pending indefinitely after the direct child exits; there is no bounded fallback.
- `child.killed` also makes repeat termination a no-op (abort after a stdout-cap termination does nothing).

**To Reproduce**

Run the equivalent of the qmd regression test in `test/datasource/skills/obsidian.test.ts` ("reaps a qmd descendant that keeps the inherited stdio open after the direct child exits") against `DiscrawlClient` with a fake `binaryPath` that spawns a child with `stdio: "inherit"` and then exits: the promise never settles on the unfixed client.

**Expected behavior**

The client settles within a bounded time even when a pipe holder survives the direct child, and reaps that holder.

**Suggested fix**

Reuse `src/process/terminate-tree.ts` (`terminateProcessTree`, added in #1582) and mirror the `exit`-path reap plus `TREE_KILL_GRACE_MS` settle guard from `src/datasource/skills/obsidian/client.ts`, with a regression test at the `DiscrawlClient` seam.

**Additional context**

Tracked separately from the katok (missing `detached`) and mailcrawl (SIGTERM-only, same missing guard) gaps. AutoRAG-side: discrawl ships as a single Mach-O binary via Homebrew, so no upstream change is required.

 - OS: macOS 26.0 (darwin arm64)
 - Node 26.8.2 / Bun 1.3.14
 - discrawl 0.15.1 via Homebrew at /opt/homebrew/bin/discrawl

lorenzozanee:
Fixes #1585

## Summary

The discrawl client could remain pending when a direct child exited while a descendant still held inherited stdout or stderr open. This change reaps the process tree on exit and bounds settlement with the shared grace period, so the client settles even when a pipe holder survives the direct child.

## Verification

- `bunx vitest run test/datasource/skills/discrawl/client.test.ts`
- `bunx vitest run test/datasource/skills/obsidian.test.ts`
- `bun run typecheck`
- `bun run build`
- `bunx biome check src/datasource/skills/discrawl/client.ts test/datasource/skills/discrawl/client.test.ts`
- `git diff --check`

Upstream CI and Windows taskkill behavior were not run locally.


vkehfdl1 (approved):
재리뷰 완료: 최신 main 기준 diff와 discrawl 관련 테스트, lint, typecheck, build를 확인했습니다.

PR #1627, opened 2026-09-19
fix(cli): surface MinSync result and failure state in refresh envelope (#1611)

### LINKED ISSUE #1611: [BUG] `autorag refresh --json` omits the MinSync result — a failed semantic sync stays invisible
## Summary
`autorag refresh --json` never reports the MinSync result: the envelope contains parsed counts, diagnostics and datasources, but no `minsync` field, even though `AutoRAGRefreshResult` carries one. A failed or skipped semantic sync is therefore invisible in the documented verification path.

## Repro
Fresh workspace (single `.txt` corpus, model verified):
```bash
autorag refresh --method parsed,minsync --json
```
Observed:
```json
{ "ok": true, "counts": { "scanned": 1, "written": 1, "deleted": 0, "skipped": 0 }, "diagnostics": [] }
```
MinSync did run — `.autorag/minsync/.minsync/cursor.json` exists and `minsync status --format json` reports `UpToDate` — but nothing in the JSON says so.

Live example of the hidden failure: on my workspace, while the MinSync embedder was failing (`minsync check --format json` -> `embedder_ok: false`, `"embedding failed: retries exhausted ... TEI API error 502 Bad Gateway: {\"error\":\"Embedding upstream request failed.\",\"code\":\"upstream\"}"`, `index_state.unindexed_rows: 34`), the concurrent `autorag refresh` still finished with `"ok": true` and no diagnostic about the failure.

## Root cause
`refreshEnvelope` (`src/cli/output.ts:33-58`) only emits `ok`, `counts`, `diagnostics` and `datasources`:
```ts
const envelope: Record<string, unknown> = { ok: true, counts: {...}, diagnostics: [...] };
if (result.datasources && result.datasources.length > 0) { envelope.datasources = ...; }
return envelope;
```
`AutoRAGRefreshResult` (`src/agent/agent.ts:145`) defines `minsync?: AutoRAGMinSyncRefreshResult`, and datasource results are surfaced while the MinSync component is not.

## Impact
- The documented check (`docs/minsync-setup.md`: "A normal setup is: ... `autorag refresh --method parsed,minsync --json`") cannot confirm semantic indexing; the only way to check is to go find `cursor.json` (the AGENTS.md live-QA gate has to do exactly that).
- Cron/watch automation cannot distinguish "synced" from "embedding failed", so semantic indexing can be silently broken until a user notices bad search results. Related closed report: #1366.

## Expected
Include the MinSync result (state, counts/chunks, cursor/manifest info, diagnostics) in the JSON envelope, and report a non-`ok`/diagnostic when the semantic sync fails.

Environment: AutoRAG `main` @ `964b3138` (post-v2.5.0), macOS arm64 (Darwin 25.6.0), Node 26.8.2, MinSync 0.4.4, profile `qwen3-embedding-0.6b` (1024d, local gateway).

vkehfdl1:
## Summary
Fixes #1611.

Previously, `autorag refresh --json` completely omitted the `minsync` component result from the JSON envelope (`refreshEnvelope`), and hardcoded `"ok": true`. If the MinSync embedder or preflight check failed (e.g. TEI 502 Bad Gateway), the refresh command silently reported success without surfacing the failure or diagnostics.

## Changes
- **Surface `minsync` result in `refreshEnvelope` & `renderRefreshHuman`**: include `ok`, `synced`, `reason`, and diagnostics in the refresh envelope when MinSync is executed.
- **Compute envelope `ok` accurately**: `envelope.ok` evaluates to `false` when MinSync or any datasource fails, or when error diagnostics are present.
- **Attach diagnostics to MinSync failures**: `MinSyncClient.sync()` now attaches diagnostic descriptors on preflight check and sync failures; `AutoRAGAgent.refresh()` maps them into `result.diagnostics` and `getRefreshStatus()`.
- **Preserve path opacity**: sanitize error reasons and diagnostic messages with `sanitizeDiagnosticMessage` so local paths never leak.
- **Unit & Integration Tests**:
  - `test/cli/output.test.ts`: test `renderRefresh` JSON envelope and human rendering for successful, failed, and path-sanitized MinSync runs.
  - `test/cli/commands.refresh-status.test.ts`: assert `minsync` presence in envelope and `ok: false` with diagnostics on embedder preflight failure.
  - `test/agent/refresh-status.test.ts`: verify failure diagnostics propagate to `result` and `getRefreshStatus()`.


vkehfdl1:
## Review

The envelope change is the right fix for #1611: `refresh --json` now carries the MinSync result and stops reporting `ok: true` over a semantic index that never updated. Three problems in the original push, all fixed in `774f5909` on this branch:

**1. The CLI test never used its fake binary (CI red on build + macos).**
`src/cli/config.ts` deliberately ignores a persisted `minSync.binaryPath` (`MINSYNC_LEGACY_IGNORED`), so `writeConfig({ binaryPath: fake })` was a no-op and MinSync was resolved from PATH. On a dev box with a real `minsync` installed the test passed with `check-failed`; on CI, with no `minsync`, it got `missing-binary`. The test now owns PATH with a fixture directory, which is deterministic either way (verified locally both with and without `/Users/jeffrey/.cargo/bin/minsync` on PATH).

**2. `#!/bin/sh` fixtures are not spawnable on Windows (CI red on windows).**
`spawn EFTYPE` in `src/minsync/process.ts`. Both fixtures are now Node scripts, matching the existing `fake-minsync.mjs` pattern in `test/minsync/minsync.test.ts` that `portableSpawnCommand` already handles on Windows. The CLI test skips on win32 (PATH injection of a shebang fixture is POSIX-only); the same failure path stays covered cross-platform by the agent test.

**3. `sanitizeDiagnosticMessage` was duplicated verbatim in `src/agent/agent.ts` and `src/cli/output.ts`,** and the renderer copy rewrote *every* diagnostic message — including datasource ones. That regex eats any slash-hierarchical string, so a source-native identity such as `/kakao/<instance>/chunks/<chunk>` would have been rendered as `<path>`, which is exactly the traceability the datasource contract wants to keep. Sanitization now happens once, where raw MinSync text enters the public result (the agent), and the CLI renderer is a pure projection again.

Also folded in:
- `MinSyncDiagnosticCode` gains `sync-failed`. A failed `minsync check`/`sync` spawn was being reported as `embedder-unavailable`, which sends users to `autorag models prefetch` for a failure that may have nothing to do with the embedder. Only an embedder preflight failure keeps that code.
- The duplicated `ok` derivation and the duplicated minsync-diagnostic derivation are each one helper now.
- README documents the `minsync` block and the `ok` semantics.

## Verification

- `biome check .` clean (only the 4 pre-existing warnings that main also has).
- `tsc --noEmit`, full suite (147 test files via `scripts/run-test-files.mjs`), and `bun run build`: all green locally.
- Live E2E cold path (`runner.mjs live --mode cold`), local gateway + `qwen3-embedding-0.6b`, `OPENAI_API_KEY` unset: `exitCode 0`, minsync `ready`, cursor written, semantic hit resolved to an absolute readable corpus source, lanes 4 PASS / 7 SKIP / 0 FAIL (skips are unregistered harnesses / no native store, same as the existing baseline).

Approving once CI is green.

PR #1630, opened 2026-09-19
test(cli): isolate MinSync PATH and autoInstall in refresh-status CLI tests

vkehfdl1:
## Problem

On `main`, the macOS compatibility CI job failed with a 60s vitest timeout in `test/cli/commands.refresh-status.test.ts` (`refresh then status emits JSON with counts and no leaked paths`, elapsed 60156ms).

### Root Cause
1. `writeConfig()` in `test/cli/commands.refresh-status.test.ts` omitted `minSync`, which defaulted to `{ enabled: true, autoInstall: true }` under config normalization.
2. In macOS CI (where MinSync is not installed on PATH), `runRefresh()` attempted auto-install / MinSync execution, which under the new native embedding setup downloaded assets/models and took > 60s, triggering vitest timeout.
3. On developer machines with `minsync` on PATH, `beforeEach` did not strip ambient MinSync binaries from `process.env.PATH`, allowing ambient binaries to leak into the test environment.

## Solution
1. Strip any host `minsync` / `minsync.exe` binaries from `process.env.PATH` in `beforeEach` using a portable `pathWithoutMinsync()` helper.
2. Default `writeConfig()` in `test/cli/commands.refresh-status.test.ts` to `minSync: { autoInstall: false }` to keep integration tests hermetic and network-free per project discipline, while preserving explicit test overrides.

## Verification
- Local bun test: `bun test test/cli/commands.refresh-status.test.ts` (9/9 passed in 546ms, down from timeout/hang)
- Full test runner: `node scripts/run-test-files.mjs` (passed)
- Biome check & typecheck: clean


vkehfdl1:
Review result: do not merge yet. PR #1630 fixes MinSync isolation in commands.refresh-status.test.ts, but test/cli/commands.watch.test.ts has the same inherited PATH/default autoInstall setup. macOS CI run 35439890028 fails because watch --once times out after 60s during MinSync auto-install/embedding. Please apply equivalent isolation to the watch test and rerun CI.

PR #1648, opened 2026-09-20
feat(agent): per-datasource search tools so one connection can be searched without fan-out

vkehfdl1:
## Problem

`search_datasource_documents` always fans out: `ParallelRetriever` spawns every authorized datasource CLI and the `scope` argument only filters results afterwards. An agent that already knows which connection holds the answer — from `check_memory` hints or from the question itself — still pays for and receives every other datasource's hits, and datasources without the `scoped` capability (katok) leak through the scope filter entirely. There was no way to search only Discord, or only the `kakao-work` account.

## Change

Generate one tool per authorized datasource connection: `search_datasource_<id>` (e.g. `search_datasource_discord`, `search_datasource_kakao_work` for a `kakao-work` account alias; ids are sanitized to tool-safe names).

- Each tool routes to a new `AutoRAGAgent.searchSingleDatasourceDocuments(datasourceId, query, { topK?, scope? })`, which registers only that connection's retrieval methods with the retriever — **no other datasource CLI is spawned**, unlike scope narrowing, which filters after execution.
- Tools are built from the same trusted config + access context as the datasource skill list: a datasource disabled in `config.json` or denied by `datasourceAccess` never appears in the tool list, and an unauthorized `datasourceId` yields an empty result set (default-deny preserved).
- Generated names join the reserved tool names, count toward the search-tool budget and weak-signal memory (`SEARCH_TOOLS` becomes an instance set), and render in the system prompt as `search only the <id> datasource connection` instead of "caller-provided tool".
- `search_datasource_documents` remains for deliberate cross-datasource fan-out; its description now points at the dedicated tools.

Multi-account connections are separate config entries (aliases), so each account gets its own tool automatically. Channels inside one connection stay scope-narrowable within its tool rather than exploding the tool list.

## Verification

- `bunx vitest run` — **1702 passed**, 5 skipped, 0 failed (full suite).
- `bun run lint` — clean (only the 4 pre-existing warnings also present on main).
- `bun run typecheck` — one pre-existing error (`src/web/search/browser-loader.ts` cannot find `puppeteer-core`), reproduces on clean main.

## Tests

`test/agent/search-single-datasource-tool.test.ts`:

- tool-name sanitization (`kakao-work` → `search_datasource_kakao_work`)
- tool execution routes to the provider with the spec's datasource id and params
- empty query short-circuits without touching the provider
- agent-level: executing one connection's tool runs only that datasource's methods (spy on the sibling stays at 0)
- unauthorized datasource id returns an empty result set without executing anything
- system prompt renders generated tools with their connection, not as caller-provided tools


vkehfdl1 (commented):
Reviewed (agent review, repo-local verification):

- Per-connection tools are generated only from trusted config + access context at construction; disabled/denied datasources never appear as tools. Init order and duplicate-name first-wins handling are safe.
- searchSingleDatasourceDocuments matches searchDatasourceDocuments exactly for filter/memory/budget handling (verified side by side).
- typecheck clean; lint shows no new warnings (4 pre-existing warnings in p2p files, also on main); test/agent 27 files / 176 tests pass, including the new search-single-datasource-tool tests.

No bugs found. Checks all green; merge is blocked only by the required review gate.

PR #1649, opened 2026-09-20
docs(datasource): steer skill manifests to dedicated search tools and document native CLIs

vkehfdl1:
## Problem

Every datasource skill manifest told the agent to "Call `search_datasource_documents`" — the fan-out tool that spawns every datasource CLI — and none documented the datasource's native CLI, even though the datasource contract expects agents to be able to drive the CLIs directly through bash. With the per-connection `search_datasource_<id>` tools from #1648, the manifests were steering agents to the wrong tool and leaving the direct-CLI path undocumented.

## Change

Every built-in datasource skill manifest now:

- names its dedicated tool (`search_datasource_kakao`, `search_datasource_discord`, `search_datasource_slack`, …) and says **do not use `search_datasource_documents` for this datasource** — it fans out to every datasource CLI;
- carries a `## Native CLI` section with copy-pastable invocations taken from each client wrapper: `katok search bm25 … --json`, `discrawl --json search --mode hybrid …`, `mailcrawl search --mode hybrid … --json`, `slacrawl --json search -workspace <teamId> …`, `notcrawl search …` (TSV, no `--json`), `wacrawl --sync never search …`, `telecrawl search …`, `clawgallery search --json --mode …`, `qmd` with its managed-environment caveat, and connector-backed notes for `rclone`, `mdfind`, and `gh`;
- reminds that datasource virtual paths (`/kakao/...`) are not OS paths and must not go to bash.

Supporting changes:

- The canonical tool-name rule moves to `src/datasource/tool-naming.ts` (`datasourceSearchToolName`) so manifests can name their tool without importing the agent layer; the agent's `singleDatasourceToolName` re-exports it.
- `AliasedDatasourceSkill` rewrites the tool reference inside manifests, so a `kakao-work` connection's manifest names `search_datasource_kakao_work`, not the template's tool.
- Crawler definitions gain `nativeCliSearchExample` (per-CLI flags differ); connector definitions gain an optional `nativeCliNote` (default: no external search CLI, the dedicated tool is the path).
- `AGENTS.md` tool table documents the generated per-connection tools; `docs/datasource-skills.md` now points at the dedicated tool and permits read-only `rclone` inspection per the manifest (credentials stay with rclone).

Stacks on #1648 (the manifests name the tools that PR generates).

## Verification

- `bunx vitest run` — full suite green (see checks).
- `bun run lint` — clean (only the 4 pre-existing warnings also on main).
- `bun run typecheck` — one pre-existing `puppeteer-core` error, reproduces on clean main.
- Rendered manifests inspected for aliased katok (`kakao-work`), slack, and cloud-drive connections.

## Tests

`test/datasource/skill-manifest-tools.test.ts` builds all 14 built-in skills through the config factory and asserts every manifest names its dedicated tool, forbids the fan-out tool, and has a Native CLI section; plus an alias case asserting the rewritten tool name.


vkehfdl1 (commented):
Reviewed (agent review). Found and fixed one bug:

- katok manifest Native CLI example said `katok context <chunkId> --json`, but katok has no top-level context command — fixed to `katok chunk context <chunkId> --json` in def44735, verified against the installed katok CLI.
- All other manifest CLI examples (discrawl, mailcrawl, qmd, wacrawl, telecrawl, slacrawl, notcrawl, clawgallery) verified against the locally installed CLIs' --help output.
- Alias manifest tool-name rewrite (kakao -> kakao-work) verified; typecheck clean; full suite 168 files / 1705 tests pass.

